PR #7, opened 2022-07-01
finished the test_if_else

aepaysinger:


tests/hacker_rank/test_if_else.py
```diff
@@ -21,6 +21,22 @@ def test_weird_stuff_weird():
 
     assert capturedOutput.getvalue() == "Weird\n", f"Printed: {capturedOutput.getvalue()}, instead of Weird"
 
+def test_weird_stuff_weird():
```
W-Ely, line 24: This function and the one below have the same name as one above.  Naming function the same within the same module isn't allowed in Python.  When you ran this, it likely didn't actually run all the tests.  Did you count the dots?

W-Ely (changes requested):
Rename functions.

W-Ely (approved):

PR #10, opened 2022-07-05
added HR division

aepaysinger:


code_challenges/hacker_rank/division.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import division
```
W-Ely, line 1: `from __future__ import division` shouldn't be needed in python 3

W-Ely (changes requested):
^

W-Ely (approved):

PR #12, opened 2022-07-06
finishing write a function

aepaysinger:


tests/hacker_rank/test_write_a_function.py
```diff
@@ -0,0 +1,22 @@
+import io
+import sys
```
W-Ely, line 2: These imports are unused, so should be removed.

W-Ely (changes requested):
^

W-Ely (approved):

PR #13, opened 2022-07-08
Hr print function

aepaysinger:


tests/hacker_rank/test_print_function.py
```diff
@@ -0,0 +1,19 @@
+from code_challenges.hacker_rank.print_function import count_up
+
+
+def test_count_up_4():
+    actual = count_up(4)
+
+    assert actual == '1234', f"Printed: {actual}, instead of 1234"
```
W-Ely, line 7: This function doesn't print anything, so the assert message is a little misleading.

W-Ely (approved):

PR #15, opened 2022-07-09
finished HR compress the string

aepaysinger:


code_challenges/hacker_rank/compress_the_string.py
```diff
@@ -0,0 +1,18 @@
...
+        else:
+            print(f'({count}, {current_letter}) ', end = '')
+            current_letter = letter
+            count = 1
+print(f'({count}, {current_letter})', end = '')
+    
+
+if __name__ == '__main__':
+    a_string = "aabbbcaafff"
+    # print(a_string)
```
W-Ely, line 17: Remove unused code.

code_challenges/hacker_rank/compress_the_string.py
```diff
@@ -0,0 +1,18 @@
...
+    current_letter = a_string[0]
+    
+    for letter in a_string:
+        if letter == current_letter:
+            count += 1
+        else:
+            print(f'({count}, {current_letter}) ', end = '')
+            current_letter = letter
+            count = 1
+print(f'({count}, {current_letter})', end = '')
```
W-Ely, line 12: This global print statement should either be in the if name is main block or removed.

W-Ely (changes requested):
From this one forward, no more prints in your functions.  If the hacker rank say 'print a thing' construct a string within and return it from the function.  The print statement should be in the `if __name__ == "__main__":` block.

Then your tests will all check for `return` values.

W-Ely (approved):

PR #18, opened 2022-07-11
Hr find the runner up score

aepaysinger:


W-Ely (changes requested):
^

tests/hacker_rank/test_find_the_runner_up_score.py
```diff
@@ -1,4 +1,5 @@
 from code_challenges.hacker_rank.find_the_runner_up_score import runner_up
+from code_challenges.hacker_rank.find_the_runner_up_score import check_tie
```
W-Ely, line 2: This is okay, but another more common way to write this is:
```
from code_challenges.hacker_rank.find_the_runner_up_score import check_tie, runner_up
```

Also while [PEP-008](https://peps.python.org/pep-0008/#imports) doesn't have an opinion on this, you'll find that by convention these should be alphabetized.  So the order here `check_tie, runner_up` matters.

W-Ely (approved):

PR #20, opened 2022-07-15
Hr nested lists

aepaysinger:


code_challenges/hacker_rank/nested_lists.py
```diff
@@ -0,0 +1,32 @@
+def student_grades(records):
+
+    lowest_score = 100
+    second_lowest_score = 100
+    lowest_students = []
+    second_lowest_students = []
+    for student_name_and_score in records:
```
W-Ely, line 7: Instead of `for student_name_and_score in records:` how about expanding it out so the code below is a bit easyer to read with `for name, score in records:`

code_challenges/hacker_rank/nested_lists.py
```diff
@@ -0,0 +1,32 @@
...
+            second_lowest_students = lowest_students
+            lowest_students = [student_name_and_score[0]]
+        elif student_name_and_score[1] == lowest_score:
+            lowest_students.append(student_name_and_score[0])
+        elif student_name_and_score[1] < second_lowest_score:
+            second_lowest_students = [student_name_and_score[0]]
+            second_lowest_score = student_name_and_score[1]
+        elif student_name_and_score[1] == second_lowest_score:
+            second_lowest_students.append(student_name_and_score[0])
+    if second_lowest_students == []:
```
W-Ely, line 20: This is good but can be simplified with `if not second_lowest_students :`

code_challenges/hacker_rank/nested_lists.py
```diff
@@ -0,0 +1,32 @@
...
+        return "No Second Lowest"
+    else:
+        return "\n".join([name for name in sorted((second_lowest_students))])
```
W-Ely, line 23: What all in `return "\n".join([name for name in sorted((second_lowest_students))])
Write` is needed?

Did you try `"\n".join(sorted(second_lowest_students))`?

W-Ely (approved):

PR #22, opened 2022-07-20
Hr designer door mat

aepaysinger:


aepaysinger:
i know that i have print statements and i will fix that later but i have been working on this for a while and i am stuck. i have drawn the top part of the design but im not sure how to reverse what i have done. original problem : https://www.hackerrank.com/challenges/designer-door-mat/problem?isFullScreen=true
what im using to help : https://www.hackerrank.com/challenges/text-alignment/problem

W-Ely:
Does this help?
```python
In [11]: detail = ".|."

In [12]: result = []

In [13]: for n in range(1,6,2):
    ...:     result.append(detail * n)
    ...:

In [14]: result.append("some text")

In [15]: for n in range(5,0,-2):
    ...:     result.append(detail * n)
    ...:

In [16]: print("\n".join(result))
.|.
.|..|..|.
.|..|..|..|..|.
some text
.|..|..|..|..|.
.|..|..|.
.|.

In [17]:```

**Comment by aepaysinger** [comment]
i have finished this and now i am working on the test cases. there are specific parameters for the size of door mat :Mat size must be N * M . (N is an odd natural number, and M is 3 times N .) I want to make a test case with even numbers and have it return a message that says it wont work due to the rules. im not exactly sure how to do that though. im also working on the other test cases

**Comment by W-Ely** [comment]
> I want to make a test case with even numbers and have it return a message that says it wont work due to the rules. im not exactly sure how to do that though.

In your function `draw_door_mat` you need to check first thing to see if `N` is not odd, then `return` what you want or `raise` the error you want. Then you can check this in your tests.

File: tests/hacker_rank/test_designer_door_mat.py [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+------.|..|..|..|..|..|..|.------\n---.|..|..|..|..|..|..|..|..|.---\n---------------HEY---------------\n\
+---.|..|..|..|..|..|..|..|..|.---\n------.|..|..|..|..|..|..|.------\n---------.|..|..|..|..|.---------\n\
+------------.|..|..|.------------\n---------------.|.---------------"
+
+    assert actual == expected, f"Returned {actual} instead of {expected}"
+
+
+def test_door_mat_even_greeting():
+    actual = draw_door_mat(["7", "21"], "HOLA")
+    expected = "---------.|.---------\n------.|..|..|.------\n---.|..|..|..|..|.---\n\
```
W-Ely, line 34: another way of writing this that ia a bit more readable would be:
```python
    expected = (
        "---------.|.---------\n"
        "------.|..|..|.------\n"
        "---.|..|..|..|..|.---\n"
        "---------HOLA--------\n"
        "---.|..|..|..|..|.---\n"
        "------.|..|..|.------\n"
        "---------.|.---------"
    )
```
Notice there are no `+` signs here.  Python will infer these should be concatenated and do so.

code_challenges/hacker_rank/designer_door_mat.py
```diff
@@ -0,0 +1,55 @@
...
+    length = int(size[1])
+    starter = int((length - 1) / 2)
+    greeting_design_count = int((length - int(len(greeting))) / 2)
+    detail_a = ".|."
+    detail_b = "-"
+    result = []
+    detail_b_count = int((length - 1) / 2)
+    detail_a_count = 1
+
+    if height % 2 != 0:
```
W-Ely, line 12: Here instead of nesting all the logic inside the `guard clause`. This should probably be like this for readability:

```
    if height % 2 == 0:
        return (
            "Mat size must be N * M . (N is an odd natural number, and M is 3 times N."
        )
        
    for n in range((int(height / 2))):
        result.append(.....
````

[Reading on guard clauses.](https://medium.com/lemon-code/guard-clauses-3bc0cd96a2d3)

File: code_challenges/hacker_rank/designer_door_mat.py [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+        else:
+            result.append(
+                (detail_b * greeting_design_count)
+                + greeting
+                + (detail_b * greeting_design_count)
+            )
+
+        detail_b_count = int(length / height)
+
+        for n in range((int(height / 2))):
```
W-Ely, line 36: Same here `range((int(height / 2)))`.

tests/hacker_rank/test_designer_door_mat.py
aepaysinger: okay. i see how that reads so much better. I was using the "\" to make it not so long but i like this way better. Thanks!

aepaysinger: the back slash

W-Ely:
Last bit of clean up is to remove any unused variables.

W-Ely (approved):

PR #23, opened 2022-07-27
Hr exceptions

aepaysinger:


code_challenges/hacker_rank/exceptions.py
```diff
@@ -0,0 +1,25 @@
+def find_the_exception(number_of_test_cases, the_values):
```
W-Ely, line 1: `number_of_test_cases` isn't used.

code_challenges/hacker_rank/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+        except ValueError:
+            if first.isdigit():
+                result.append(
+                    f"Error Code: invalid literal for int() with base 10: '{second}'"
+                )
+            else:
+                result.append(
+                    f"Error Code: invalid literal for int() with base 10: '{first}'"
+                )
+    return "\n".join(map(str, result))
```
W-Ely, line 17: What is map doing for you here?  Is there a different approach that would work without using map?

aepaysinger: i want to print out a list on separate lines, and the items can sometimes be an int.
i cant use just join on a none string
what i did to change it was make the int a string after the division.
i could also do a loop but im not sure how to return all the results in a loop. if i have the return statement at the end of a for loop it will just exit after the first iteration

code_challenges/hacker_rank/exceptions.py
```diff
@@ -0,0 +1,24 @@
+def find_the_exception(the_values):
+    result = []
+    for first, second in the_values:
+        try:
+            result.append(str(int(int(first) / int(second))))
```
W-Ely, line 5: Casting to string here is what I had in mind. 


In `str(int(int(first) / int(second)))` if the intention is to drop decimal places with the surrounding `int`.  Would integer division work ( `//`) instead?

W-Ely (commented):
This just needs the tests fixed.

W-Ely (approved):

PR #24, opened 2022-08-01
Hr company logo

aepaysinger:


code_challenges/hacker_rank/company_logo.py
```diff
@@ -0,0 +1,31 @@
+def make_logo(name):
+    letters = {}
+    for letter in name:
+        if letter == " ":
+            letters[letter] = 0
+            continue
```
W-Ely, line 6: If an empty space isn't a valid character to count, why not skip it rather than adding it to the dict at all?

Might add a test here to ensure it is doing what you expect.  For example `aa bb cc dd` should return `a 2\nb 2\nc 2` even though there are `3` space characters.  The code does this, but the test will prove it.

W-Ely (approved):

W-Ely (approved):

PR #26, opened 2022-08-06
working on

aepaysinger:


code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
+import random
```
W-Ely, line 1: The file name should be lower_snake_case.py.

aepaysinger: fixed.

code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for i in range(int(input("enter how many students you have: "))):
+        student_names.append(input("Enter student name: "))
+
+    for name in student_names:
+        student_id_number = random.randint(111111, 999999)
+        student_id_numbers.append(student_id_number)
+        while student_id_number in student_id_numbers:
+            student_id_number = random.randint(111111, 999999)
+        student_name = name.split(" ")
```
W-Ely, line 18:
```
for name in student_names:
    student_id_number = random.randint(111111, 999999)
    student_id_numbers.append(student_id_number)  # <--- When would you want this to happen?
    while student_id_number in student_id_numbers:
        student_id_number = random.randint(111111, 999999)
    student_name = name.split(" ")
```
I think after you are sure it isn’t already in the list.  So probably something like:
```
ffor name in student_names:
    student_id_number = random.randint(111111, 999999)
    while student_id_number in student_id_numbers:
        student_id_number = random.randint(111111, 999999)
    student_id_numbers.append(student_id_number) 
    student_name = name.split(" ")
```

code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
...
+        if len(student_name) > 2:
+            email_name.append(student_name[0][0] + student_name[1][0] + student_name[2])
+        else:
+            email_name.append(student_name[0][0] + student_name[1])
```
W-Ely, line 22: If you're handling the 3 long case it might make sense to handle the 1 long case too.

code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for i in range(len(student_names)):
+        new_id = student_id_numbers[i]
```
W-Ely, line 25: Instead of `new_id` here `student_id_number` would be better.

code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
...
+        new_id = str(new_id)[3:]
```
W-Ely, line 26: same

code_challenges/Account_Generator.py
```diff
@@ -0,0 +1,60 @@
...
+        student_email_addresses.append(f"{email_name[i]}{new_id}@example.com")
+
+    for i in range(len(student_names)):
+        return f"name: {student_names[i]}\nid: {student_id_numbers[i]}\nemail: {student_email_addresses[i]}\n"
```
W-Ely, line 30: This isn't doing what you think.

aepaysinger: probably not. i had it as a print statement until i sent it to you and quickly changed it without really thinking about it so thank you!

code_challenges/Account_Generator.py
aepaysinger: right that makes sense. the name matters.

code_challenges/Account_Generator.py
aepaysinger: what do you mean by that?

W-Ely: `if len(student_name) > 2:` is true for `"William Ely Paysinger".split(" ")`
else would happen for both `"William Paysinger".split(" ")` or `"Ely".split(" ")`

code_challenges/hacker_rank/rps.2.py
```diff
@@ -0,0 +1,41 @@
+import random
+
+
+def choose_rps():
+    "output: randomly returns rock, paper, or scissors"
+    r = random.randint(0,2)
```
W-Ely, line 6: Consider: `random.randchoice(...)`

PR #27, opened 2022-08-11
Cw youre a square

aepaysinger:


W-Ely (dismissed):
Probably this is meant to check for _perfect_ squares.

You have three code paths and only 2 tests.  What would the last test look like?

aepaysinger:
it would be for less than 0

code_challenges/code_wars/youre_a_square.py
```diff
@@ -14,4 +14,6 @@ def is_square(n):
 if __name__ == "__main__":
     n = 34
     print(is_square(n))
-    
\ No newline at end of file
+    
+
+    # fun things
```
W-Ely: Stray comment

W-Ely (changes requested):
I didn't read the whole commit.  I see now it is a very long message.

aepaysinger:
i think this is cleaned up now

W-Ely (approved):

PR #32, opened 2022-08-17
Cw sort array

aepaysinger:


code_challenges/code_wars/sort_array.py
```diff
@@ -0,0 +1,9 @@
+def sort_array(numbers_to_sort):
+
+    odds = iter(sorted(number for number in numbers_to_sort if number % 2))
+    return [next(odds) if number % 2 else number for number in numbers_to_sort]
```
W-Ely, line 4: This kind of solution isn't allowed unless you have spent the time to fully understand and remember it, which would be great.

What does `sorted` do?
What does `iter` do?
What does `next` do?

Also how does `odds` wind up being only odds?

Why do the even numbers stay put and the odds not?

W-Ely: [sorted](https://docs.python.org/3/library/functions.html#sorted) takes an iterable and returns a sorted list.  What arguments are you passing to sorted? `number for number in numbers_to_sort if number % 2`, this is a [Generator Expression](https://peps.python.org/pep-0289/) which is iterable, but not something you have used before.  It's similar to [List Comprehensions](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) but a different thing, no list is created.

`iter` and `next` can be used when `__iter__` and `__next__` are defined on the object you are using or in this case since the they are used on a list the [C language equivalent](https://github.com/python/cpython/blob/main/Objects/listobject.c).  [iter doesn't loop through the list](https://github.com/python/cpython/blob/main/Objects/listobject.c#L3183), it sets up an iterator adding `it_index` = 0 to the object.  Then each time `next` is called on the object, it returns that index and [increments it_index](https://github.com/python/cpython/blob/main/Objects/listobject.c#L3230) so the next `next` call will pull the next index's item.

W-Ely (approved):

PR #34, opened 2022-08-17
finished

aepaysinger:


code_challenges/code_wars/find_short.py
```diff
@@ -0,0 +1,15 @@
+def find_short(s):
+    shortest = ""
+    seperated_words = s.split(" ")
+
+    for word in seperated_words:
+        if shortest == "":
+            shortest += word
+        if len(word) < len(shortest):
+            shortest = word
+    return len(shortest)
```
W-Ely, line 10: This works, but the += is a little odd.  What you really want is to just reassign. like `shortest = word` since shortest isn't ever accumulated and should always only be a single word.  It works since it starts as just `""`.

Also see how this compares:
```python
def find_short(s):
    seperated_words = s.split(" ")
    shortest = seperated_words[0] if seperated_words else ""
    for word in seperated_words:
        if len(word) < len(shortest):
            shortest = word
    return len(shortest)
```

This syntax `shortest = seperated_words[0] if seperated_words else ""` is called a [ternary operators](https://book.pythontips.com/en/latest/ternary_operators.html#ternary-operators)

W-Ely (approved):

code_challenges/code_wars/find_short.py
aepaysinger: okay. i relooked at it. do i even need the if statement? doesn't it work just as well to just set shortest to the first string?

W-Ely (approved):

code_challenges/code_wars/find_short.py
aepaysinger: interesting. I made a test with an empty string and it returned 0

W-Ely: I see, yes if `s` is even empty it will create a list with at least an empty string in the first place.

PR #36, opened 2022-08-22
Cw cakes

aepaysinger:


code_challenges/code_wars/cakes.py
```diff
@@ -0,0 +1,40 @@
+def how_many_cakes(recipe, available):
+    if len(available) < len(recipe):
+        return 0
+
+    amount_of_cakes_per_ingredients = {}
+    for key in recipe:
+        if key not in available:
+            pass
+        else:
+            amount_of_cakes_per_ingredients[key] = available[key] // recipe[key]
```
W-Ely, line 10: Instead of doing nothing on the fist condition can you change it so that it's the one that does the thing and then you don't need the else.

W-Ely (approved):

code_challenges/code_wars/cakes.py
aepaysinger: when i do this i get a key error on my test with no matching ingredients.

W-Ely: Any time you have a condition that just `pass`es you don't need it.
```python
if key in available:
    amount_of_cakes_per_ingredients[key] = available[key] // recipe[key]
```

README.md
```diff
@@ -1 +1,11 @@
 # code_challenges
+## CodeWars
+### Cakes (7kyu)
+   * Module: `code_challenges/code_wars/cakes.py`
+   * Tests:`tests/code_wars/test_cakes.py`
+   * URL: [challenge url](https://www.codewars.com/kata/525c65e51bf619685c000059/python)
+  ---
```
W-Ely, line 7: I'd remove these '---' , but it's only personal preference. I don't readily see a reason this section should be broken up, but as you like.

W-Ely (changes requested):
Commented out code and readme doesn't belong in this PR I think.

W-Ely (approved):

PR #41, opened 2022-08-30
CW-highest scoring word

aepaysinger:


code_challenges/code_wars/highest_scoring_word.py
```diff
@@ -0,0 +1,51 @@
...
+            word_points += letter_points[letter]
+        score_board[word] = word_points
+    highest_word = None
+    highest_score = 0
+    for key in score_board:
+        if score_board[key] > highest_score:
+            highest_score = score_board[key]
+            highest_word = key
+        elif score_board[key] == highest_score:
+            highest_word += " " + key
```
W-Ely, line 44: `for word in score_board:` would be a little more descriptive here.

code_challenges/code_wars/highest_scoring_word.py
```diff
@@ -0,0 +1,51 @@
+def high(words):
```
W-Ely, line 1: `highest_scoring_word(words):` would be a better name for this function.

tests/code_wars/test_highest_scoring_word.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_high_nultiple_same_score():
+    actual = high("ade j ee gc")
+    expected = "ade j ee gc"
+
+    assert actual == expected, f"Returned {actual} instead of {expected}"
+
+
+def test_high():
```
W-Ely, line 18: This function is named the same as the top one. This isn't allowed. Are you sure both are running?

W-Ely (changes requested):
Only the test using the same name is a required change.  Others are just suggestions.

W-Ely (approved):

PR #44, opened 2022-09-07
completed sum of the smallest

aepaysinger:


tests/adhoc/test_sum_of_the_smallest.py
```diff
@@ -0,0 +1,29 @@
...
+    actual = sum_of_the_smallest([5, 1, 1, 9, 3, 2])
+    expected = 2
+
+    assert actual == expected, f"Returned {actual} instead of {expected}"
+
+
+def test_sum_of_the_smallest_multiple_same():
+    actual = sum_of_the_smallest([-5, -1, -1, -9, -3, -2])
+    expected = None
+
```
W-Ely, line 28: Unfortunately I didn't catch this in person, but this function is named the same as the one before :-/

PR #45, opened 2022-09-07
CW redo the list

aepaysinger:


code_challenges/code_wars/redo_the_list.py
```diff
@@ -0,0 +1,18 @@
...
+    length = len(s)
+
+    for i, number in enumerate(s):
+        if i % 2 == 0:
+            t.append(s[i])
+            t.append(s[-1 - i])
+        else:
+            t.append(s[-1 - i])
+            t.append(s[i])
+    del t[length:]
```
W-Ely, line 12: `del t[length:]` nice.

This essentially means you are doing twice the work you need to.  Since only have the indices on the first half of the input list are needed and then rest is discarded.  Can you think of a way to not do this extra work?

W-Ely (approved):

code_challenges/code_wars/redo_the_list.py
aepaysinger: i couldn't thats why i ended up using del. i tried thinking of a while loop but couldn't manage to get it to work.

PR #47, opened 2022-09-08
Adhock

aepaysinger:
completed matrix and sum of most frequent number and tests for both

### DIFF
--- a/code_challenges/adhoc/rotate_v2.py
+++ b/code_challenges/adhoc/rotate_v2.py
@@ -1,8 +1,0 @@
-def rotate(matrix):
-    rotated = []
-    for i in range(len(matrix)):
-        row = []
-        for j in range(len(matrix) - 1, -1, -1):
-            row.append(matrix[j][i])
-        rotated.append(row)
-    return rotated
--- a/code_challenges/adhoc/rotate_v2.py
+++ b/code_challenges/adhoc/rotate_v2.py
@@ -1,0 +1,8 @@
+def rotate(matrix):
+    rotated = []
+    for i in range(len(matrix)):
+        row = []
+        for j in range(len(matrix) - 1, -1, -1):
+            row.append(matrix[j][i])
+        rotated.append(row)
+    return rotated
--- a/code_challenges/adhoc/rotate_matrix.py
+++ b/code_challenges/adhoc/rotate_matrix.py
@@ -1,0 +1,28 @@
+def rotate_matrix(matrix):
+    rotated_matrix = []
+    length = 0
+    for path in matrix:
+        length += 1
+    count = 0
+    while count < (length * length):
+        for path in reversed(matrix):
+            rotated_matrix.append(path[0])
+            del path[0]
+            count += 1
+    final_matrix = []
+    start = 0
+    stop = length * length
+    step = length
+    for i in range(start, stop, step):
+        final_matrix.append(rotated_matrix[i:i+step])     
+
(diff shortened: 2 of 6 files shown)


W-Ely (changes requested):
This is an innovative solution, good job!

A couple other approaches I'd like you to study and understand:
1) nested for loops, the most 'expected' answer.
```python
def rotate(matrix):
    rotated = []
    for i in range(len(matrix)):
        row = []
        for j in range(len(matrix) - 1, -1, -1):
            row.append(matrix[j][i])
        rotated.append(row)
    return rotated
```

2) A solution using [zip](https://docs.python.org/3.3/library/functions.html#zip):
```python
def rotate(matrix):
    return [list(r) for r in zip(*matrix[::-1])]
```

Add these 2 solutions to the file as rotate_v2 and rotate_v3 along with tests. Then in the PR provide a description of how the are working. For example in solution 2:
- What is this syntax called `[var for var in iterable]`?
- What does `zip()` do, with an example? 
- What does the `*` do on matrix?
- What does [::-1] do?
- Why is `list(r)` needed?

code_challenges/adhoc/rotate_v2.py
```diff
@@ -0,0 +1,8 @@
+def rotate(matrix):
+    rotated = []
+    for i in range(len(matrix)):
+        row = []
+        for j in range(len(matrix) - 1, -1, -1):
+            row.append(matrix[j][i])
+        rotated.append(row)
+    return rotated
```
W-Ely, line 8: close, add these functions in the same file as the other.  So instead of new files named rotate_v2, you will have a new function named rotate_v2 in the same file as the original.

W-Ely:
> along with tests

😅

code_challenges/adhoc/rotate_v2.py
W-Ely: Also - have to remove the other files rotate_v2 and rotate_v3 since they are now functions in this file, and add the same tests that are for the other function for these functions too.

W-Ely:
Let me know with a comment once you thing all the feed back has been addressed.

aepaysinger:
def rotate_v3:
matrix[::-1] returns the matrix backwards 
the * takes it out of the list. so instead of a list of lists you just have lists
then list r for r in
takes each number a the same position from each list and puts them in a list
then because it is list comprehension all of them are in a list

aepaysinger:
def rotate_v2:
the only part i still dont understand is the stop at -1

W-Ely:
> matrix[::-1] returns the matrix backwards

Yup.
>the * takes it out of the list. so instead of a list of lists you just have lists

More or less, this is called un-packing https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists 

> then list r for r in
takes each number a the same position from each list and puts them in a list

This isn't correct, or the explanation is missing information.  Can you create an example that omits `zip` to illustrate your conclusion.


> then because it is list comprehension all of them are in a list

Yup.


You left out `zip`.  What _exactly_ does zip return?


---------------

> the only part i still dont understand is the stop at -1

Start, Stop, Step.  What's the starting number? What's stoping number? And what does `step` actually mean?

aepaysinger:
i think we talked about this but I thought I would comment on it so you know I know. starting number is len(range) - 1, stop is -1 (because you want 0 to run but nothing before that), and step is how you loop through the list, so if its -1 your are going backwards one by one

W-Ely (approved):

PR #48, opened 2022-09-12
Cw consecutive strings

aepaysinger:
still not passing the codewars, but i think they are wrong this time. so i have no completed a readme


W-Ely (dismissed):
Some reading:
[sorted](https://docs.python.org/3/howto/sorting.html) including the [key functions](https://docs.python.org/3/howto/sorting.html#key-functions) block.
[lambda](https://docs.python.org/3/reference/expressions.html#lambda) also called anonymous functions, functions without a name or `def`

After reading the above see if the below makes sense.

What if instead of storing only the words in the word tracker dict lists you stored both the word and it's position. For the list `["a", "bb", "cc", "ddd", "ee"]`
```
In [5]: l = ["a", "bb", "cc", "ddd", "ee"]

In [6]: d = {}
   ...: for i, s in enumerate(l):
   ...:     if d.get(len(s)) == None:
   ...:         d[len(s)] = [(s, i)]
   ...:     else:
   ...:         d[len(s)] += [(s, i)]
   ...:

In [7]: d
Out[7]: {1: [('a', 0)], 2: [('bb', 1), ('cc', 2), ('ee', 4)], 3: [('ddd', 3)]}
```

Then add to the list in sorted order:
```
In [10]: k = 3

In [11]: w = []

In [12]: for key in sorted(d, reverse=True):
    ...:     if len(w) >= k:
    ...:         break
    ...:     w += d[key]
    ...:

In [13]: w
Out[13]: [('ddd', 3), ('bb', 1), ('cc', 2), ('ee', 4)]

In [14]:
```

Then finally take the k words in order.
```
In [19]: final = []

In [20]: for word, _ in sorted(w, key=lambda word_and_position: word_and_position[1]):
    ...:     if len(final) == 3:
    ...:         break
    ...:     final.append(word)
    ...:

In [21]: "".join(final)
Out[21]: 'bbccddd'
```

W-Ely:
Lot's of extra files here :-[

aepaysinger:
okay. they werent green on the side. but i went through the files changed and have seen them and will update accordingly

aepaysinger:
for the files that are empty do i just need to do a git rm?

W-Ely:
> for the files that are empty do i just need to do a git rm?

You could try that.  But also ask yourself how they got there and try to not do the thing.  

If rm doesn't work and the file is empty, why not delete it?

code_challenges/code_wars/consecutive_strings.py
```diff
@@ -0,0 +1,19 @@
+def consecutive_strings(names, k):
+    if k > len(names):
+        return ""
+    length = 0
+    concatenated_string = ""
+    for i in range(0, k, 1):
```
W-Ely, line 6: `for i in range(0, k, 1):` is not doing the correct thing here. Once you figure out why it's broken, write a test that shows it's fixed.

W-Ely (changes requested):

code_challenges/code_wars/consecutive_strings.py
aepaysinger: i think i fixed it. let me know what you think and i wrote my last test for it.

W-Ely (dismissed):

code_challenges/code_wars/sort_array.py
```diff
@@ -1,9 +0,0 @@
-def sort_array(numbers_to_sort):
```
W-Ely, line 1: Why are you deleting this file?

aepaysinger: im not sure why it ended up here but i already have a brach pushed through onto main of sort array

W-Ely: If you delete this file and commit the deletion, when you merge it will be removed from main.

This doesn't seem like you intention.   

1) checkout to main.
2) open this file in vs code and add a single space, don't save but leave the file open
3) check back out to this branch
4) remove the added space from this file and save the file back on this branch.  (Now the file should look exactly like it looks on main.
5) Add and commit this file
6) Push the changes to this branch on Github.

If all goes well, the file will be removed from the diff, since it's already on main.

PR #51, opened 2022-09-20
CW scramble

aepaysinger:
i pass all the tests in codewars but ran into a runtime error, so i have not completed the README yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,8 @@
   * Module: `code_challenges/code_wars/help_the_bookseller.py`
   * Tests: `tests/code_wars/test_help_the_bookseller.py`
   * URL: [challenge url](https://www.codewars.com/kata/54dc6f5a224c26032800005c/python)
+### Scramblies (5kyu)
+  * Module: `code_challenges/code_wars/scramble.py`
+  * Tests: `tests/code_wars/test_scramble.py`
+  * URL: [challenge url](https://www.codewars.com/kata/55c04b4cc56a697bb0000048/python)
+  * Run time: On
--- a/README.md
+++ b/README.md
@@ -48,4 +48,4 @@
   * Module: `code_challenges/code_wars/scramble.py`
   * Tests: `tests/code_wars/test_scramble.py`
   * URL: [challenge url](https://www.codewars.com/kata/55c04b4cc56a697bb0000048/python)
-  * Run time: On
+  * Run time: O(n)
--- a/code_challenges/code_wars/scramble.py
+++ b/code_challenges/code_wars/scramble.py
@@ -1,0 +1,19 @@
+def scramble(s1, s2):
+    matching_word = unscramble(s1, s2)
+ 
+    return s2 == matching_word
+
+
+def unscramble(s1, s2):
+    matching_word = ""
+    s1 = list(s1)
+    for letter in s2:
+        if letter in s1:
+            matching_word += letter
+            s1.remove(letter)
+    return matching_word
+
+if __name__ == "__main__":
+    s1 = 'dmweyrrnbovczlge'
+    s2 = 'dywegzwbvnm'
+    print(scramble(s1, s2))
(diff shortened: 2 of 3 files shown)


code_challenges/code_wars/scramble.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def unscramble(s1, s2):
+    matching_word = ""
+    s1 = list(s1)
+    for letter in s2:
+        if letter in s1:
+            matching_word += letter
+            s1.remove(letter)
+    return matching_word
```
W-Ely, line 14: `s1 = list(s1)` - Python has to do on thing for each character in the string
`for letter in s2` - Python has to do a thing for each item in the list:
      that thing - `if letter in s1:`. How does python know if it's in the list?  It has to check each item in the list.
      also in the loop - `s1.remove(letter)` Think about the steps that have to be done by python to remove an item.  It has to first find it by checking every item until it gets there.  It then has to take it out, but what does taking it out look like?  What does it have to do to correct all the indices? 
      
Probably you can see that if the strings are very long at all you're asking python to do a whole lot of things.  How can you reduce them?

W-Ely (commented):
A Runtime error is a specific thing https://runestone.academy/ns/books/published/fopp/Debugging/RuntimeErrors.html 

Your code takes too long to run.  So you would say something like 'The execution times out.'  Or 'It doesn't complete in the required time.'

README.md
```diff
@@ -44,3 +44,8 @@
   * Module: `code_challenges/code_wars/help_the_bookseller.py`
   * Tests: `tests/code_wars/test_help_the_bookseller.py`
   * URL: [challenge url](https://www.codewars.com/kata/54dc6f5a224c26032800005c/python)
+### Scramblies (5kyu)
+  * Module: `code_challenges/code_wars/scramble.py`
+  * Tests: `tests/code_wars/test_scramble.py`
+  * URL: [challenge url](https://www.codewars.com/kata/55c04b4cc56a697bb0000048/python)
+  * Run time: On
```
W-Ely, line 51: This is written `O(n)`

tests/code_wars/test_scramble.py
```diff
@@ -1,4 +1,4 @@
-from code_challenges.code_wars.scramble import scramble, unscramble
+from code_challenges.code_wars.scramble import count_characters, scramble
```
W-Ely, line 1: Which of these imports is used.  Only the ones used should be imported.

Also, write unit tests for each function you write, not just the outside function.

So in this case you do need the import since the new tests you write will need it.

W-Ely (changes requested):
Need more tests, each function you write should be tested.

Condition practice:
When you have code like
```python
for character in counter2:
    if character not in counter1:
        return False
    if counter2[character] > counter1[character]:
        return False
```
Consider the if statements.  There are 2 that both return False.  Combine them into one if statement.

tests/code_wars/test_scramble.py
```diff
@@ -1,15 +1,35 @@
...
-    expected = True
+def test_count_characters():
+    actual = count_characters("rkqodlw", "world")
+    expected = {"r": 1, "k": 1, "q": 1, "o": 1, "d": 1, "l": 1, "w": 1}, {
+        "w": 1,
+        "o": 1,
+        "r": 1,
+        "l": 1,
+        "d": 1,
+    }
```
W-Ely, line 12: Info only, no need to change:  If you wanted these to format better you could write this as:
```python
    expected = (
        {"r": 1, "k": 1, "q": 1, "o": 1, "d": 1, "l": 1, "w": 1},
        {"w": 1, "o": 1, "r": 1, "l": 1, "d": 1},
    )
```

W-Ely (approved):
🍠

tests/code_wars/test_scramble.py
aepaysinger: great info cuz thats how i wanted to do it and wasn't sure how! thank you!!!

PR #52, opened 2022-09-23
Cw weight for weight

aepaysinger:


README.md
```diff
@@ -44,3 +44,8 @@
   * Module: `code_challenges/code_wars/help_the_bookseller.py`
   * Tests: `tests/code_wars/test_help_the_bookseller.py`
   * URL: [challenge url](https://www.codewars.com/kata/54dc6f5a224c26032800005c/python)
+### Weight for weight (5kyu)
+  * Module: `code_challenges/code_wars/weight_for_weight.py`
+  * Tests: `tests/code_wars/test_weight_for_weight.py`
+  * URL: [challenge url](https://www.codewars.com/kata/55c6126177c9441a570000cc/python)
+  * Run time: On2
```
W-Ely, line 51: This would be written `O(n²)` or `O(n^2)` but isn't actually correct in this case.

It is true that when you see a nested loop you should think carefully about the code since it is likely to be n², but in this case what is it actually doing?

Consider this code:
```python
l = ["a", "b", "c"]
for a in l:
    for b in l:
        print(a, b)
```
The input is 3 long, but there are 9 prints. 3^2 == 9 so clearly n².

But in the w_for_w function:
```python
w = [1, 1, 2, 3, 4, 4, 5, 6]
i = 0
j = 1
while i < len(w):
    print("i is", i)
    while j < len(w) and w[i] == w[j]:
        print("j is", j)
        j += 1
    if len(w[i:j]) > 1:
        print("slice is", w[i:j])
    i = j
    j = i + 1
```
Something like:
```
i is 0
j is 1
slice is [1, 1]
i is 2
i is 3
i is 4
j is 5
slice is [4, 4]
i is 6
i is 7
```
So this piece of code while looking quite complex is still O(n).

But...  that isn't the full story story here.  You are using the Python's builtin `sorted`. What is its time complexity?  It uses the [Timsort](https://en.wikipedia.org/wiki/Timsort) algorithm.  <-- read about it and find the big O.

W-Ely (changes requested):

README.md
W-Ely: > But... that isn't the full story story here. You are using the Python's builtin sorted. What is its time complexity? It uses the [Timsort](https://en.wikipedia.org/wiki/Timsort) algorithm. <-- read about it and find the big O.

Did you miss this in the above? I see you've changed to O(n). Where in the wiki lead you to believe sort is O(n)?

aepaysinger: would it be O(log n)?
from what i have gather thus far, the Timsort algorithm is a combination of insertion sort and merge sort which are O(log n).

W-Ely: https://en.wikipedia.org/wiki/Timsort#Analysis `O(n log n)`

aepaysinger: that was a typo i have it written down as O(n log n). im still watching some videos to understand this all more. I just watched one on binary insertion sort

W-Ely (approved):

PR #53, opened 2022-09-26
adhoc - class counter

aepaysinger:


code_challenges/adhoc/counter.py
```diff
@@ -0,0 +1,15 @@
+# create a counter that has an incriment , decremet, and value method (return current value)
+class Counter:
+    def __init__(self):
+        self._value = 0
+        pass
```
W-Ely, line 5: Pass was here to start because it was need to make an empty function syntactically correct. 

You can't have a ":" at the end of a line without code of some kind in the block below.  It serves no purpose now.

code_challenges/adhoc/counter.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def value(self):
+        value = self._value
+        return value
```
W-Ely, line 9: When you find yourself assigning a variable only to return it, you don't need the assignment.

code_challenges/adhoc/counter.py
```diff
@@ -0,0 +1,14 @@
+# create a counter that has an incriment , decremet, and value method (return current value)
+class Counter:
+    def __init__(self):
+        self._value = 0
+        
+
```
W-Ely, line 6: Per pip8  single line between functions within a class https://peps.python.org/pep-0008/#blank-lines

W-Ely (approved):

PR #54, opened 2022-09-28
Cw trigrams

aepaysinger:


code_challenges/code_wars/trigrams.py
```diff
@@ -0,0 +1,17 @@
+def trigrams(phrase):
+    i = 0
+    j = 3
+    trigrams = []
+    updated_phrase = phrase.replace(" ", "_")
+    while i < len(phrase):
```
W-Ely, line 6: What is this `while` loop doing?

code_challenges/code_wars/trigrams.py
```diff
@@ -0,0 +1,17 @@
...
+        for character in updated_phrase:
+            
+            trigrams.append(updated_phrase[i:j])
+            i += 1
+            j += 1
+
+    return " ".join([trigram for trigram in trigrams if len(trigram) == 3])
+
```
W-Ely, line 14: Per pep-008 https://peps.python.org/pep-0008/#blank-lines 2 blank lines surrounding functions in the global scope

code_challenges/code_wars/trigrams.py
```diff
@@ -0,0 +1,17 @@
...
+    trigrams = []
+    updated_phrase = phrase.replace(" ", "_")
+    while i < len(phrase):
+        for character in updated_phrase:
+            
+            trigrams.append(updated_phrase[i:j])
+            i += 1
+            j += 1
+
+    return " ".join([trigram for trigram in trigrams if len(trigram) == 3])
```
W-Ely, line 13: How could you avoid needing to do this inner list comprehension?

W-Ely (changes requested):

code_challenges/code_wars/trigrams.py
```diff
@@ -0,0 +1,18 @@
...
+    trigrams = []
+    updated_phrase = phrase.replace(" ", "_")
+    while i < len(phrase):
+        for character in updated_phrase:
+            print(i, j, updated_phrase[i:j])
+            trigrams.append(updated_phrase[i:j])
+            i += 1
+            j += 1
+
+    return " ".join(trigrams[:-2])
```
W-Ely, line 13: What if you were to move this slice to line 7, so you never do the work to begin with rather than throwing it away at the end? 

Also you don't use `character` from line 7 so what if you changed this to a range and made it output the correct `i` value.  You then wouldn't need to increment `i`.

If i and j are always the same distance apart do you need to increment a j to begin with or could you use i for both?

code_challenges/code_wars/trigrams.py
```diff
@@ -0,0 +1,17 @@
...
+    trigrams = []
+
+    updated_phrase = phrase.replace(" ", "_")
+    
+    for i in range(len(updated_phrase)):
+        if len(updated_phrase[i:j]) == 3:
+            trigrams.append(updated_phrase[i:j])
+            j += 1
+        else:
+            return " ".join(trigrams)
```
W-Ely, line 12: Now that you are iterating on a range, how about not including the full length since you want to stop `j` short of the end?  Then you wont need the if check.

Something like:
```python
def trigrams(phrase):
    chars_per_ngram = 3
    ngrams = []
    phrase = phrase.replace(" ", "_")
    for i in range(len(phrase) - chars_per_ngram + 1):
        ngrams.append(phrase[i:i + chars_per_ngram])
    return " ".join(ngrams)
```

aepaysinger: i was taking 2 off the end but range(len(phrase) - Chars_per_ngram + 1) is stopping before i get to the last 2

W-Ely (approved):

PR #55, opened 2022-09-29
Cw first non repeating letter

aepaysinger:
this was on the CW youre a square branch. so i removed it from there and put it on the right branch.

### DIFF
--- a/code_challenges/code_wars/first_non_repeating_letter.py
+++ b/code_challenges/code_wars/first_non_repeating_letter.py
@@ -1,10 +1,7 @@
 def first_non_repeating_letter(word):
     word = list(word)
     for i, letter in enumerate(word):
-        if letter in word[i + 1 :: 1] or letter in word[0:i]:
-            print(letter, word[i + 1 :: 1], word[0:i])
-            continue
-        else:
+        if letter not in word[i + 1 :: 1] or letter not in word[0:i]:
             return letter
 
 
--- a/code_challenges/code_wars/first_non_repeating_letter.py
+++ b/code_challenges/code_wars/first_non_repeating_letter.py
@@ -1,12 +1,13 @@
 def first_non_repeating_letter(word):
-    for letter in word:
-        if word.find(letter):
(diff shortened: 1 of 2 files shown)


code_challenges/code_wars/first_non_repeating_letter.py
```diff
@@ -0,0 +1,13 @@
+def first_non_repeating_letter(word):
+    word = list(word)
+    for i, letter in enumerate(word):
```
W-Ely, line 3: a string is both enumerate-able and sliceable so don't think you need to convert it to a list first

code_challenges/code_wars/first_non_repeating_letter.py
```diff
@@ -0,0 +1,13 @@
...
+        if letter in word[i + 1 :: 1] or letter in word[0:i]:
+            print(letter, word[i + 1 :: 1], word[0:i])
+            continue
```
W-Ely, line 6: Do you remember what I said about if blocks that do nothing?

W-Ely (approved):
Is there a readme?  What's the time complexity of this solution?

How might you reduce the time complexity here?

code_challenges/code_wars/first_non_repeating_letter.py
```diff
@@ -0,0 +1,12 @@
+from collections import Counter
+
+
+def first_non_repeating_letter(word):
+    for letter in word:
+        if Counter(word)[letter] == 1:
+            return letter
```
W-Ely, line 7: What does `Counter(word)` do?  Does the case of the import tell you what kind of Python thing Counter is?

[docs: counter](https://docs.python.org/3/library/collections.html#collections.Counter)

Use counter but don't create a new one for every letter, it's an object so reuse the instance you create a single time.

aepaysinger: i think i fixed it. creating a variable, and then using that.

W-Ely (approved):

PR #56, opened 2022-09-29
substring

aepaysinger:


W-Ely (approved):
Great job, now read about [str.find](https://docs.python.org/3/library/stdtypes.html#str.find)

PR #57, opened 2022-10-03
Cw song

aepaysinger:


tests/code_wars/test_song.py
```diff
@@ -0,0 +1,8 @@
+from code_challenges.code_wars.song import Song
+
+
+def test_song():
+    mount_moose = Song("Mount Moose", "The Snazzy Moose")
+
```
W-Ely, line 6: Should probably test that `mount_moose.title` and `moose.artist` are what you would expect also.

code_challenges/code_wars/song.py
```diff
@@ -0,0 +1,15 @@
+class Song:
+    def __init__(self, title, artist):
+        self.title = title
+        self.artist = artist
+        self.names_of_listeners = set()
```
W-Ely, line 5: This is a good time to talk about the concept of public vs private attributes.

When you create a new instance with `my_instance = Song("a", "b")`, this instance has methods and properties that are available for usage, all of which are attributes.  Sometimes we need to assign a property to an instance for internal use and we don't want it to be a part of the externally available properties.  For example in this case `title`, `artist` and `how_many` are all expected to be `public`, but `names_of_listeners` isn't supposed to be used outside the instance. Maybe it's okay if it is, but maybe you want to reserve the right to later change the way this is implemented, how it is named for example, without breaking peoples code that might use this class.

To indicate you do not recommend usage outside this class, you prefix the name of the property with an `_`. So `self.names_of_listeners = set()` should become `self._names_of_listeners = set()`.  Then when others are reading your code, they will see the underscore and know that this is an `implementation detail` and shouldn't trust it to remain consistent.  Nothing about the `_` forces that it can't be used, it's just a name, but by convention everyone will know.

W-Ely (changes requested):

code_challenges/code_wars/song.py
aepaysinger: so you use the "_" to show that it should only be used inside this class?

W-Ely: > so you use the "_" to show that it should only be used inside this class?

Yup.

tests/code_wars/test_song.py
W-Ely: Bump

W-Ely (approved):

PR #60, opened 2022-10-12
Cw car class

aepaysinger:


README.md
```diff
@@ -69,7 +69,12 @@
   * Tests: `tests/code_wars/test_person.py`
   * URL: [Challenge url](https://www.codewars.com/kata/513f887e484edf3eb3000001/python)
   * Run Time: O(1)
+### Instance (6kyu)
+  * Module `code_challenges/code_wars/instance.py`
+  * Tests: `tests/code_wars/test_instance.py`
+  * URL: [Challenge url](https://www.codewars.com/kata/561f9d37e4786544e0000035/python)
+  * Run Time: O(1)
```
W-Ely, line 76: It's really O(n log n) can you spot why?

code_challenges/code_wars/instance.py
```diff
@@ -0,0 +1,21 @@
+class Vehicle:
+    def __init__(self, seats, wheels, engine):
+        self.seats = seats
+        self.wheels = wheels
+        self.engine = engine
+
+
+class Planet:
+    def __init__(self, moon):
+        self.moon = moon
```
W-Ely, line 10: These aren't really a part of the solution, but part of testing, so should live in the test file instead.

W-Ely (changes requested):

README.md
aepaysinger: is it because the if block cuts the time in half?

PR #61, opened 2022-10-14
CW delete extra

aepaysinger:


code_challenges/code_wars/delete_extra.py
```diff
@@ -0,0 +1,17 @@
+def delete_extra(alist, amount):
+    updated = []
+    count = {}
+    for item in alist:
+        if item not in count:
+            count[item] = 1
+        else:
+            count[item] += 1
```
W-Ely, line 8: How could you refactor this to use the `dict.get` method so you don't need the if/else check?

W-Ely (changes requested):

code_challenges/code_wars/delete_extra.py
aepaysinger: i'm not sure. wouldn't count.get return the value?

W-Ely: > wouldn't count.get return the value?

It does, but also returns a default value when the key isn't present.

Consider:
```python
count[item] = count.get(item, 0) + 1
```

`.get(<key>, <optional default>)`. If no default is provided, like .get("the key"), the default value when the item isn't present is `None`.  This wouldn't be helpful here, but `0` is.

Moving forward if you consider using syntax like `if item not in count:` I'd like you to try and use `.get` instead.  Not that one is exactly better, but it is an important part of the language.

PR #62, opened 2022-10-18
Cw football cards

aepaysinger:


code_challenges/code_wars/football_cards.py
```diff
@@ -0,0 +1,55 @@
+def football_cards(team_cards):
+    a_players = 11
+    b_players = 11
+    a_cards = {}
+    b_cards = {}
+    while a_players > 6 and b_players > 6:
+        for card in team_cards:
+            if a_players == 6 or b_players == 6:
+                break
```
W-Ely, line 9:
```
    while a_players > 6 and b_players > 6:
        for card in team_cards:
            if a_players == 6 or b_players == 6:
                break
```
Do you need this while loop at all?

W-Ely: You should read up on `re` the [Python built in](https://docs.python.org/3/library/re.html) for matching and finding things.

An example implementation:
```
import re

team_cards = ["A4R", "A6R", "A8R", "A10R", "A11R"]
for card in team_cards:
    match = re.match(r"([AB])([0-9]+)([YR])", card)
    team, player, penalty = match.groups()
    print(team, player, penalty)
```

While regex/regular expressions can be a lot to think about and write, this is the perfect case were it would make the code a lot more readable.

code_challenges/code_wars/football_cards.py
```diff
@@ -0,0 +1,55 @@
...
+                        a_players -= 1
+                else:
+                    player_number = card[1] + card[2]
+                    a_cards[player_number] = a_cards.get(player_number, "") + card[3]
+                    if (
+                        a_cards[player_number] == "R"
+                        or a_cards[player_number] == "YY"
+                        or a_cards[player_number] == "YR"
+                    ):
+                        a_players -= 1
```
W-Ely, line 28: Great job!  Getting it working is always the first goal.  Now lets think about how to make it a little better.

The logic above in the if block is exactly duplicated below in the else block except with different variable names.  Can you think of a way to [DRY (don't repeat yourself)](https://en.wikipedia.org/wiki/Don't_repeat_yourself) this code up some?

Consider your variables:
```python
a_players = 11
b_players = 11
a_cards = {}
b_cards = {}
```
How could your code change if you were to use a structure like:
```python
teams = {
    "A": {"players": 11, "cards": {}},
    "B": {"players": 11, "cards": {}},
}
```

code_challenges/code_wars/football_cards.py
aepaysinger: i tried getting rid of the while loop earlier and the code didnt work. which i didnt understand. but i just re did it and it does still work. i must have missed pulling something back when i did it before.

README.md
```diff
@@ -78,8 +78,12 @@
   * Module: `code_challenges/code_wars/vector_class.py`
   * Tests: `tests/code_wars/test_vector_class.py`
   * Run Time: n/a
+### Football Cards (6kyu)
+  * Module: `code_challenges/code_wars/football_cards`
+  * Tests: `tests/code_wars/test_football_cards`
+  * Run Time: O(n log n)
```
W-Ely, line 84: What lead you to this Big O choice?

aepaysinger: i cant remember. i was looking up how to figure it out and something i read made me go this route. after re reading i am thinking it is O(n). there is one for loop and im not doing any sorting.

W-Ely (approved):

PR #63, opened 2022-10-19
Cw count smileys

aepaysinger:
to make into into 2 functions i could make create the face into a function then call that function in the main function to get the dictionary of face elements.

### DIFF
--- a/code_challenges/code_wars/count_smileys.py
+++ b/code_challenges/code_wars/count_smileys.py
@@ -1,0 +1,20 @@
+def count_smileys(smiley_faces):
+    parts_of_face = {"eyes": [":", ";"], "nose": ["-", "~"], "mouth": ["D", ")"]}
+    count = 0
+    for face in smiley_faces:
+        if len(face) == 2:
+            if face[0] in parts_of_face["eyes"]:
+                if face[-1] in parts_of_face["mouth"]:
+                    count += 1
+        if len(face) == 3:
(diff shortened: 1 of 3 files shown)


code_challenges/code_wars/count_smileys.py
```diff
@@ -0,0 +1,20 @@
...
+            if face[0] in parts_of_face["eyes"]:
+                if face[-1] in parts_of_face["mouth"]:
+                    count += 1
+        if len(face) == 3:
+            if face[0] in parts_of_face["eyes"]:
+                if face[-1] in parts_of_face["mouth"]:
+                    if face[1] in parts_of_face["nose"]:
+                        count += 1
+
+    return count
```
W-Ely, line 15: This is good, consider the following.  No need to change yours, just wanting to show a different approach.

```python
def count_smileys(smiley_faces):
    allowed_eyes, allowed_noses, allowed_mouths = {":", ";"}, {"-", "~"}, {"D", ")"}
    count = 0
    for face in smiley_faces:
        eyes = face[0]
        nose = face[1] if len(face) == 3 else None
        mouth = face[2] if len(face) == 3 else face[1]
        if (
            eyes in allowed_eyes and
            (not nose or nose in allowed_noses) and 
            mouth in allowed_mouths
        ):
           count += 1
    return count
```

W-Ely (approved):

PR #64, opened 2022-10-24
CW rot13

aepaysinger:


code_challenges/code_wars/rot13.py
```diff
@@ -0,0 +1,28 @@
...
+                    code += chr(count)
+            else:
+                if count > 90:
+                    new = count - 90
+                    code += chr(64 + new)
+                else:
+                    code += chr(count)
+        else:
+            code += character
+    return code
```
W-Ely, line 20: Great!  Good job exploring and learning about chr and ord!

13, 122, 96, 90, and 64 are what are called `magic numbers`.  These are numbers that are sprinkled into your code that a reader might not understand and easy to miss if you need to change a thing.

Something to keep in mind for next time: A better way of handling this would be to assign these numbers into a variable that describes them and is then reusable. like
```python
    rotate_count = 13
    upper_bound = 122
    lower_bound = 90
    upper_start = 96
    lower_start = 64
```

Take a look at this solution, which simplifies the nested if/else and reduces code duplication:
```python
def rot13(a_string):
    rotated_string = ""
    rotate_count = 13
    upper_bound = 122
    lower_bound = 90
    upper_start = 96
    lower_start = 64
    for character in a_string:
        if not character.isalpha():
            rotated_string += character
            continue

        bound = upper_bound if character.islower() else lower_bound
        start = upper_start if character.islower() else lower_start

        new_ord = ord(character) + rotate_count
        if new_ord > bound:
            new_ord = new_ord - bound
            rotated_string += chr(start + new_ord)
        else:
            rotated_string += chr(new_ord)

    return rotated_string
```

W-Ely (approved):

code_challenges/code_wars/rot13.py
```diff
@@ -1,18 +1,23 @@
 def rot13(a_string):
     code = ""
+    amount_to_shift = 13
+    upper_alpha_min = 65
+    upper_alpha_max = 90
+    lower_alpha_min = 97
```
W-Ely, line 6: The only times you use the mins you add one to them. It would be better to include that in the variable to start with.

W-Ely (approved):

W-Ely (approved):

PR #65, opened 2022-10-27
Cw maximum subarray sum

aepaysinger:
i got the code to work, but it takes too long to run.......


code_challenges/code_wars/maximum_subarray_sum.py
```diff
@@ -0,0 +1,37 @@
+def max_sequence(sequence):
+    total = 0
+    for number in sequence:
+        if number > 0:
+            total += 1
+    if total == 0:
+        return total
+    else:
+        for i, number in enumerate(sequence):
```
W-Ely, line 9: You don't ever use number here, how about using a range instead.

code_challenges/code_wars/maximum_subarray_sum.py
```diff
@@ -0,0 +1,37 @@
...
+            j = len(sequence)
+            while j > i:
+                if sequence[j-1] > 0:
```
W-Ely, line 12: What about if `sequence[i]` is negative?  If either end is negative this slice can be skipped.

code_challenges/code_wars/maximum_subarray_sum.py
```diff
@@ -0,0 +1,37 @@
...
+                    if sum(sequence[i:j]) > total:
+                        total = sum(sequence[i:j])
+                        j -= 1 
+                    else:
+                        j -= 1
+                else:
+                    j -= 1
```
W-Ely, line 19: In all three case you minus equal 1.  Only write it once outside the if block as the last thing in the while block.

W-Ely (commented):
I'll think on the time complexity issue, here are some other thoughts.

W-Ely:
Regarding time complexity. The only slices that need to be considered are the ones that terminate at a negative number. For example given sequence [1,2,3,-4,5,6,7] the only `i`s and `j`s that need to be considered are 0, 2, 4 and 6, so you could loop through the list once and create a list of the indices that need to be checked.  Rather than checking the next i when the previous was positive.

aepaysinger:
i left the code commented out at the bottom so i didnt lose what i had if i mess it up too much. this update still doesnt pass codewars. it now doesn't pass tests with large arrays to check.

W-Ely:
This is broken for a number of reasons. Consider
```python
def seq_to_check(seq):
    arrays_to_check = []
    j = len(sequence) - 1
    i = 0

    for _ in range(len(sequence)):
        while sequence[i] < 1:
            i += 1
        while sequence[j] < 1:
            j -= 1
        if i >= j:
            break
        start = i
        stop = j
        arrays_to_check.append((start, stop))
        j -= 1
        i += 1
    return arrays_to_check

actual = seq_to_check([1,2,3,-4,5,6,7])
expected = [(0,2),(0,6),(4,6)]
assert actual == expected, f"{actual} != {expected}"
```

W-Ely:
This uses 2 loops to fist find the `i`s and then find the `j`s:
```python
def seq_to_check(seq):
    starts = []
    ends = []
    previous_start = False
    previous_end = False
    for i in range(len(sequence)):
        if sequence[i] < 1:
            previous_start = False
        else:
            if not previous_start:
                starts.append(i)
            previous_start = True

    for j in range(len(sequence) - 1, -1, -1):
        if sequence[j] < 1:
            previous_end = False
        else:
            if not previous_end:
                ends.append(j)
            previous_end = True

    seq_to_check = []
    for i in starts:
        for j in ends:
            if i <= j:
                seq_to_check.append((i, j))

    return seq_to_check

actual = seq_to_check([1,2,3,-4,5,6,7])
expected = [(0,6),(0,2),(4,6)]
assert actual == expected, f"{actual} != {expected}"
```

aepaysinger:
so i changed the code and it will pass your test but i still run into issues, can we take a look at this together?

W-Ely:
For ref: https://www.codewars.com/kata/54521e9ec8e60bc4de000d6c

W-Ely:
I wrote this:
```python
def max_sequence(nums):
    if len(nums) <= 1:
        return max(0, sum(nums)) 
    max_sum = 0
    starting_points = []
    ending_points = []
    last_index = len(nums) - 1
    if nums[0] >= 0:
        starting_points.append(0)
    for i in range(1, last_index):
        if nums[i] >= 0 and nums[i - 1] < 0:
            starting_points.append(i)
        if nums[i] >= 0 and nums[i + 1] < 0:
            ending_points.append(i)
    if nums[-1] >= 0:
        ending_points.append(last_index)
        if nums[-2] < 0:
            starting_points.append(last_index)
    for i in starting_points:
        for j in ending_points:
            if j >= i:
                max_sum = max(max_sum, sum(nums[i:j + 1]))
    return max_sum
```
It works but times out.

Here is the solution ChatGPT came up with which stretches the brain on how it even works:
```python
def max_sequence(arr):
    max_sum = 0
    current_sum = 0
    
    for num in arr:
        current_sum = max(num, current_sum + num)
        max_sum = max(max_sum, current_sum)
    
    return max_sum
```

Check out the link https://chat.openai.com/share/935b1b96-ff22-4bb9-bc0e-6839acc5739c

aepaysinger:
this is brilliant!

aepaysinger:
it took me a few tries but i totally get it. max_sum is obvious but understanding the current sum took just a minute. but it is figuring out if its better to start over or keep going. how smart.

PR #66, opened 2022-10-31
CW increment string

aepaysinger:


code_challenges/code_wars/increment_string.py
```diff
@@ -0,0 +1,31 @@
+def increment_string(characters):
+    if characters == "":
+        return "1"
+    if characters[-1].isalpha():
+        characters += "1"
+    else:
```
W-Ely, line 6: Instead of nesting all the relevant code in an else block, how about adding a return statement to this guard clause.

code_challenges/code_wars/increment_string.py
```diff
@@ -0,0 +1,31 @@
...
+        numbers = ""
+        stop = None
+        for i in range(len(characters) - 1, -1, -1):
+            if characters[i].isdigit():
+                numbers += characters[i]
+                stop = i
+            else:
+                break
+
+        characters = characters[:stop]
```
W-Ely, line 16: Can you think of a way to replace this with `re` like here https://github.com/aepaysinger/code-challenges/blob/main/code_challenges/code_wars/football_cards.py#L14

code_challenges/code_wars/increment_string.py
```diff
@@ -1,31 +1,28 @@
...
-    length = len(numbers)
-    numbers = int(numbers)
+
+    parts = re.match(r"([a-zA-Z0-9]+[a-zA-Z])([0-9]+)", characters)
+
+    first, second = parts.groups()
+
+    characters = first
+
+    numbers = int(second)
```
W-Ely, line 16: Food for thought. Typically instead of doing this:
```python
first, second = parts.groups()
characters = first
numbers = int(second)
```
You would do:
```python
characters, numbers = parts.groups()
number = int(numbers)
```
because first and second are not very descriptive.

W-Ely (dismissed):

code_challenges/code_wars/increment_string.py
```diff
@@ -9,14 +9,14 @@ def increment_string(characters):
 
     parts = re.match(r"([a-zA-Z0-9]+[a-zA-Z])([0-9]+)", characters)
 
-    first, second = parts.groups()
+    first_half, nums_at_end = parts.groups()
 
-    characters = first
+    characters = first_half
```
W-Ely, line 14: Part of the issue is this reassignment only to get a better variable name.
```python
characters = first_half
```
The reason I suggested calling it characters from the jump on line 12 was to avoid this needless reassignment.

W-Ely (approved):

code_challenges/code_wars/increment_string.py
aepaysinger: i see, i guess the reason i did it this way was readability to me. i understand that the reassignment isnt needed but i like that it shows that the string that is passed into the function is getting split up. is doing it this way frowned upon?

W-Ely: It is frowned upon.  You can always pick a name that is more descriptive like `characters_first_half`

code_challenges/code_wars/increment_string.py
```diff
@@ -9,18 +9,18 @@ def increment_string(characters):
...
+    first_half_characters, nums_at_end = parts.groups()
 
-    characters = first_half
+    # characters = first_half
 
     numbers = int(nums_at_end)
     numbers += 1
     numbers = str(numbers)
     numbers = numbers.zfill(len(nums_at_end))
     numbers = str(numbers)
```
W-Ely, line 20:
```    
    numbers = int(nums_at_end)
    numbers += 1
    numbers = str(numbers)  # <--- numbers here is now a string.
    numbers = numbers.zfill(len(nums_at_end))  # <--- What python type is numbers now?
    numbers = str(numbers)  # <--- Is this line needed?
 ```

aepaysinger: no it is not

W-Ely (approved):

PR #68, opened 2022-11-09
Cw barista problem

aepaysinger:


code_challenges/code_wars/barista_problem.py
```diff
@@ -0,0 +1,17 @@
...
+        return 0
+    cleaning_time = 0
+    for i in range(1, len(orders)):
+        cleaning_time = (2 * i) + cleaning_time
+    time = 0
+    orders.sort()
+    for i in range(len(orders) + 1):
+        time += sum(orders[:i])
+
+    return time + cleaning_time
```
W-Ely, line 12: `orders.sort()` is `O(n log n)`

and sum() is O(n), but you are summing for each i in the length, so this is closer to O(n^2) even though the sum lengths are reducing. 

So this is overall O(n^@) I would say.

Also here is a helpful link on time complexity https://wiki.python.org/moin/TimeComplexity

aepaysinger: thanks for that link!

W-Ely (approved):

PR #69, opened 2022-11-11
CW valid parentheses

aepaysinger:


code_challenges/code_wars/valid_parentheses.py
```diff
@@ -0,0 +1,22 @@
...
+            count += 1
+        elif character == ")":
+            close += 1
+            count -= 1
+        if count == -1:
+            return False
+    if open == close:
+        return True
+    else:
+        return False
```
W-Ely, line 17: If you are returning what the condition evaluates to what is the better return statement here?

W-Ely (commented):
Try it with only the `count` variable.

code_challenges/code_wars/valid_parentheses.py
aepaysinger: i did try to use only count before i sent it in but ran into an issue, but i was able to fix it now you requested it  be done.

W-Ely: Try "((())"

and "(((())"

code_challenges/code_wars/valid_parentheses.py
```diff
@@ -0,0 +1,18 @@
...
+        if character == "(":
+            count += 1
+        elif character == ")":
+            count -= 1
+        if count == -1:
+            return False
+    if count >= 1:
+        return False
+
+    return True
```
W-Ely, line 13: Instead of:
```
if count >= 1:
    return False

return True
```
consider
```
return count == 0
```

W-Ely (approved):

aepaysinger:
is there anything else you would like me to change on this one?

W-Ely:
Address my comment on lines 10-13.  Either let me know you understand it with a reply or change the code to show me you understand it.

aepaysinger:
so sorry, i dont know how i missed that. after re reading my code i see how the one line is better then having the 2 return statements.

W-Ely (dismissed):

W-Ely (approved):

PR #71, opened 2022-11-22
CW sum my digits

aepaysinger:


code_challenges/code_wars/sum_my_digits.py
```diff
@@ -0,0 +1,21 @@
+def sum_my_digits(starting_number, pattern, stop):
+    digit_to_sum = str(get_number_list(starting_number, pattern, stop))
+    digit_broken_up = []
+    for digit in digit_to_sum:
+        digit_broken_up.append(int(digit))
```
W-Ely, line 5: For practice convert lines 3-5 into a list comprehension. https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions

W-Ely (changes requested):
Practice list comprehensions.

W-Ely (approved):
:sweet_potato:

PR #72, opened 2022-11-22
HR weird or not weird

aepaysinger:


code_challenges/hacker_rank/weird_or_not_weird.py
```diff
@@ -0,0 +1,16 @@
+def weird_or_not_weird(n):
+    if n % 2 != 0:
+        return "Weird"
+    elif n % 2 == 0 and n > 20:
+        return "Not Weird"
+    for i in range(2, 6):
+        if n % 2 == 0 and n == i:
+            return "Not Weird"
```
W-Ely, line 8: Instead of iterating over a range here in an outside loop, use the `in` condition like:
`if n % 2 == 0 and n in range(2,6):`

W-Ely (changes requested):

W-Ely (approved):

W-Ely (approved):

W-Ely (approved):

PR #76, opened 2022-12-07
Cw find odd int

aepaysinger:
the code wars didnt want me to use a class. so the commented section was me doing it the other way for them

### DIFF
--- a/code_challenges/code_wars/find_odd_int.py
+++ b/code_challenges/code_wars/find_odd_int.py
@@ -1,0 +1,28 @@
+class FindOddInt():
+    def __init__(self, numbers):
+        self.numbers = numbers
+        self.number_amount = {}
+
+    
+    def find_amount(self):
+        for number in self.numbers:
+            self.number_amount[number] = self.number_amount.get(number, 0) + 1
+
+        for number in self.number_amount:
+            if self.number_amount[number] % 2 != 0:
+                return number
+
+
+# def find_it(seq):
+#     number_amount = {}
+#     for number in seq:
+#         number_amount[number] = number_amount.get(number, 0) +1
+
+#     for number in number_amount:
+#         if number_amount[number] % 2 != 0:
+#             return number
+
+    
+# if __name__ == "__main__":
+#     seq = [1,1,1,1,1,1,10,1,1,1,1]
+#     print(find_it(seq))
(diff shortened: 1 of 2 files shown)


W-Ely:
> the code wars didnt want me to use a class. so the commented section was me doing it the other way for them

Code War likely requires a method as the entry point.  This doesn't mean that you can't use classes.  For example:
```python
class FindOddInt:
    def __init__(self, numbers):
        self.numbers = numbers
        self.number_amounts = {}

    def prepare_freq_map(self):
        for number in self.numbers:
            self.number_amounts[number] = self.number_amounts.get(number, 0) + 1
    
    def find_amount(self):
        for number, amount in self.number_amounts.items():
            if amount % 2 != 0:
                return number


def find_it(seq):
    find_odd_int = FindOddInt(seq)
    find_odd_int.prepare_freq_map()
    return find_odd_int.find_amount()


if __name__ == "__main__":
    seq = [1,1,1,1,1,1,10,1,1,1,1]
    print(find_it(seq))
```

code_challenges/code_wars/find_odd_int.py
```diff
@@ -0,0 +1,28 @@
+class FindOddInt():
```
W-Ely, line 1: `class FindOddInt():` --> `class FindOddInt:`

code_challenges/code_wars/find_odd_int.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, numbers):
+        self.numbers = numbers
+        self.number_amount = {}
+
```
W-Ely, line 5: extra line

code_challenges/code_wars/find_odd_int.py
```diff
@@ -0,0 +1,28 @@
+class FindOddInt():
+    def __init__(self, numbers):
+        self.numbers = numbers
+        self.number_amount = {}
```
W-Ely, line 4: Typically a collection would have a plural name.

code_challenges/code_wars/find_odd_int.py
```diff
@@ -0,0 +1,28 @@
...
+#         number_amount[number] = number_amount.get(number, 0) +1
+
+#     for number in number_amount:
+#         if number_amount[number] % 2 != 0:
+#             return number
+
+    
+# if __name__ == "__main__":
+#     seq = [1,1,1,1,1,1,10,1,1,1,1]
+#     print(find_it(seq))
```
W-Ely, line 28: Try to use the function here to call and use your class. Then it will work both places without commented out code.

W-Ely (changes requested):

PR #85, opened 2022-12-19
CW popping balloons

aepaysinger:
i created tests for the function and class and it seems to be working the way i want but for some reason it fails the codewars final test
https://www.codewars.com/kata/633b8be2b5203f003011d79e/train/python


code_challenges/code_wars/popping_balloons.py
```diff
@@ -0,0 +1,94 @@
...
+    def find_top_balloons(self):
+        top_amount = 0
+        second_amount = 0
+        top_balloon = None
+        second_balloon = None
+        
+        for balloon in self.balloons_amount:
+            if self.balloons_amount[balloon] > top_amount:
+                top_amount = self.balloons_amount[balloon]
+                top_balloon = balloon
```
W-Ely, line 23: If it's greater than the top amount, then you would have a new top amount and a new second in this same if block.

W-Ely:
> but for some reason it fails the codewars final test

Does it give any information at all on the failure?

code_challenges/code_wars/popping_balloons.py
aepaysinger: i think i have fixed this but it still fails

aepaysinger:
it just gives me the output that my code gave and what they say the output should be

W-Ely:
> it just gives me the output that my code gave and what they say the output should be

Create a test with the same input. - Might have to jerry rig your code to get this input.  For example
```python
def the_function_they_call(the_input):
    the_result = my_code(input)
    if the_result == the_result_they_dont_like:
        return input
    return the_result
```

code_challenges/code_wars/popping_balloons.py
```diff
@@ -0,0 +1,94 @@
...
+                second_balloon = balloon
+        return top_balloon, top_amount, second_balloon, second_amount
+
+    def pop_most(self):
+        top_balloon, top_amount, second_balloon, second_amount = self.find_top_balloons()
+        one_left = all(x == 1 for x in self.balloons_amount.values())
+        if one_left:
+            while len(self.popped_ballons) < self.amount:
+                self.popped_ballons.append(self.balloons[-1])
+                self.balloons.pop(-1)
```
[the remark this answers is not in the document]
aepaysinger: yes it does! i like that and can use it in a few places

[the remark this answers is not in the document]
aepaysinger: just kidding just the one spot

aepaysinger:
i dont understand how the function you wrote will give me the input

aepaysinger:
now for some reason my last two tests dont pass, but when i plug that information into the function i have the correct output so not sure what happened there

W-Ely:
> i dont understand how the function you wrote will give me the input

You said that Code Wars currently returns what you output and what they expect you to output. If this is true, you can output what they input.

Your current Code Wars entry point:
```python
def freq_stack(pops, balloons):
    balloons = Balloons(balloons)
    balloons.track_balloons()
    balloons.find_top_balloons()

    return balloons.pop_most(pops)
```
Should become:
```python
THE_OUTPUT_FROM_CODE_WARS_THEY_SAY_IS_WRONG = <copy that ouput here>

def freq_stack(pops, balloons):
    the_input = pops, balloons

    balloons = Balloons(balloons)
    balloons.track_balloons()
    balloons.find_top_balloons()
    
    the_result = balloons.pop_most(pops)
    if the_result == THE_OUTPUT_FROM_CODE_WARS_THEY_SAY_IS_WRONG:
        return the_input
    return the_result
```

aepaysinger:
still couldnt figure out how to get the output that way and would like to talk through later, but i did just put in a print statement and was able to get and feel silly it took that long.

code_challenges/code_wars/popping_balloons.py
```diff
@@ -0,0 +1,77 @@
...
+                        if self.balloons[i] == top_balloon:
+                            self.popped_ballons.append(self.balloons[i])
+                            self.balloons.pop(i)
+                            break
+            elif top_amount == second_amount:
+                the_biggest = None
+                for i in range(len(self.balloons) -1, -1, -1):
+                    if self.balloons[i] == top_balloon or self.balloons[i] == second_balloon:
+                        the_biggest = self.balloons[i]
+                        self.balloons_amount[the_biggest] -= 1
```
aepaysinger, line 51: this is the section that is having issues. i think what is happening is it reads self.balloons[i] == top_balloon first and goes through self.balloons looking for that first then goes through the list looking for second balloon. What i was wanting was for it to go through the list and look for them both at the same time and which ever appears first becomes the biggest

aepaysinger:
found and fixed my problem, for some reason my last 2 tests arent working. the expected is what i get when i put them into the function but in the test i get something else.

W-Ely (dismissed):

W-Ely (approved):

PR #91, opened 2022-12-27
Cw remove parentheses

aepaysinger:
i also moved around files in the README to be in the proper section

### DIFF
--- a/code_challenges/code_wars/remove_parentheses.py
+++ b/code_challenges/code_wars/remove_parentheses.py
@@ -1,0 +1,42 @@
+class CharacterRemover:
+    def __init__(self,  characters):
+        self.characters = characters
+
+    def item_to_remove(self):
+        self.start_item = "("
+        start_item_index = None
+        self.stop_item = ")"
+        stop_item_index = None
+        
+        for i in range(len(self.characters)):
+            if self.characters[i] == self.start_item:
+                start_item_index = i
+            elif self.characters[i] == self.stop_item:
+                stop_item_index = i
+                print(start_item_index, stop_item_index)
+                return start_item_index, stop_item_index
+
+    def remove_item(self):
+        print(self.item_to_remove)
+        start, stop = self.item_to_remove()
+        self.characters = self.characters[:start] + self.characters[stop + 1:]
+        
+        return self.characters
+
+    def final_check(self):
+        self.characters = self.remove_item()
+        if self.start_item in self.characters and self.stop_item in self.characters:
+            self.remove_item()
+        
+        return self.characters
+
+def remove_parentheses(characters):
+    characters = CharacterRemover(characters)
+    characters.item_to_remove()
+    characters.remove_item()
+    return characters.final_check()
+
+
+if __name__ == "__main__":
+    characters = "example(unwanted thing)example"
+    print(remove_parentheses(characters))
--- a/code_challenges/code_wars/remove_parentheses.py
+++ b/code_challenges/code_wars/remove_parentheses.py
@@ -19,6 +19,7 @@ class CharacterRemover:
         start, stop = self.item_to_remove()
         self.characters = self.characters[:start] + self.characters[stop + 1 :]
         if self.start_item in self.characters and self.stop_item in self.characters:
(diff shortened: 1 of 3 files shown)


code_challenges/code_wars/remove_parentheses.py
```diff
@@ -0,0 +1,36 @@
+class CharacterRemover:
+    def __init__(self, characters):
+        self.characters = characters
+
+    def item_to_remove(self):
+        self.start_item = "("
+        start_item_index = None
+        self.stop_item = ")"
+        stop_item_index = None
```
W-Ely, line 9: While Python will happily interpret this, it's generally considered best practice to define all properties that will exist in the init block even if they start out without a value.

Anything that you will attach to `self`.  So in this case `self.start_item` and  `self.stop_item`.  There are probably exceptions to this rough philosophy, but especially if the values don't change throughout the life of this instance, they should be present on init.  So since start is always `(` and end `)` moving them into the init doesn't change the way the code operates at all and brings this inline with what most would expect.

code_challenges/code_wars/remove_parentheses.py
```diff
@@ -0,0 +1,36 @@
...
+            elif self.characters[i] == self.stop_item:
+                stop_item_index = i
+                return start_item_index, stop_item_index
+
+    def remove_item(self):
+        start, stop = self.item_to_remove()
+        self.characters = self.characters[:start] + self.characters[stop + 1 :]
+        if self.start_item in self.characters and self.stop_item in self.characters:
+            self.item_to_remove()
+            self.remove_item()
```
W-Ely, line 23: This is called **recursion** and is a pretty handy way of solving some problems. It does have some draw backs though. Do you know what they are?

Try your code with the characters `"".join([f"a({n})b" for n in range(1000)])` and `"".join([f"a({n})b" for n in range(3000)])` does that work? If so, try `"ab".join([f"a({n})b" for n in range(10000)])`.

-------------------------

Is line 22 needed?

tests/code_wars/test_remove_parenrheses.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_character_remover_item_to_remove_2():
+    characters = CharacterRemover("a(b)c(d)")
+
+    assert characters.item_to_remove() == (1, 3)
+
+
+def test_character_remover_remove_item():
+    characters = CharacterRemover("this(is)it")
+    characters.item_to_remove()
```
W-Ely, line 21: Is line 21 needed?

code_challenges/code_wars/remove_parentheses.py
```diff
@@ -0,0 +1,36 @@
...
+
+        return self.characters
+
+
+def remove_parentheses(characters):
+    characters = CharacterRemover(characters)
+    characters.item_to_remove()
```
W-Ely, line 30: Is line 30 needed?

tests/code_wars/test_remove_parenrheses.py
```diff
@@ -0,0 +1,37 @@
...
+
+    assert characters.remove_item() == "thisit"
+
+
+def test_charcter_remover_remove_item_2():
+    characters = CharacterRemover("a(b)c(d)")
+    characters.item_to_remove()
```
W-Ely, line 28: Line 28?

W-Ely (changes requested):
Remove unneeded function calls.

W-Ely (dismissed):

code_challenges/code_wars/remove_parentheses.py
aepaysinger: RecursionError: maximum recursion depth exceeded while calling a Python object

aepaysinger: i thought line 22 was needed because im needing to find the new index of the item to remove

aepaysinger:
did i miss something?

W-Ely:
Yup, lines 23 and 31 are still not needed.

W-Ely:
and 22's comment should be removed.

W-Ely (approved):

PR #92, opened 2022-12-31
Cw house rental

aepaysinger:


aepaysinger:
This was meant to be a draft, sorry!
I think I have this figured out but then sometimes the result doesnt match do to a rounding situation i think. I checked the issues reported and I think it hasn't be fixed though it is stated like it is. not sure if i should make my own tests an submit to you or try to continue to get it to pass codewars. Here is the link https://www.codewars.com/kata/63aa7b499c53c5af68e86a86/python

W-Ely:
> not sure if i should make my own tests an submit to you or try to continue to get it to pass codewars. Here is the link

The goal is to write lots of Python, not so much reverse engineer someones problem because they don't clearly describe how to round for their answer.  If a challenge is included in the repo it should be a valid solution on the site.  So rather than continue this one, just delete it and move on.

aepaysinger:
removed house rental

PR #97, opened 2023-01-05
CW robot path

aepaysinger:


code_challenges/code_wars/robot_path.py
```diff
@@ -0,0 +1,61 @@
...
+        self.directions = directions
+
+    def translate_directions(self):
+        translated_instructions = []
+        direction_translation = {
+            "<": "left",
+            ">": "right",
+            "^": "up",
+            "v": "down",
+        }
```
W-Ely, line 12: This would be good to have in the `__init__` since it's always the same.

code_challenges/code_wars/robot_path.py
```diff
@@ -0,0 +1,61 @@
...
+                i = j
+                j = i + 1
+            else:
+                translated_instructions.append(
+                    f"Take {len(self.directions[i:j])} steps {direction_translation[self.directions[i]]}"
+                )
+                if j == len(self.directions):
+                    break
+                i = j
+                j = i + 1
```
W-Ely, line 49:
```python
                if j == len(self.directions):
                    break
                i = j
                j = i + 1
```
The above is in both the if and else block, can you pull it out and put it below both.

code_challenges/code_wars/robot_path.py
```diff
@@ -0,0 +1,61 @@
...
+                    return "\n".join(translated_instructions)
+            while self.directions[i] == self.directions[j]:
+                j += 1
+                if j == len(self.directions):
+                    break
+            if len(self.directions[i:j]) == 1:
+                translated_instructions.append(
+                    f"Take 1 step {direction_translation[self.directions[i]]}"
+                )
+                if j == len(self.directions):
```
W-Ely, line 38: Could the outer most for loop be changed to a while loop with this as the condition?

W-Ely:
I didn't mean add an additional while loop that isn't needed, but rather change the outer for loop to a while loop and use the conditions you are checking inside as the break out condition.  Anytime you have a for loop and always break out when a condition is met consider a while loop. Nothing wrong with the for loop, but while loops are typically harder so you should practice them.

This is what I had in mind:
```python
    def translate_directions(self):
        if self.directions == "":
            return "Paused"

        translated_instructions = []
        i, j = 0, 1

        while j < len(self.directions):
            while self.directions[i] == self.directions[j]:
                j += 1
                if j == len(self.directions):
                    break
            if len(self.directions[i:j]) == 1:
                translated_instructions.append(
                    f"Take 1 step {self.direction_translation[self.directions[i]]}"
                )
            else:
                translated_instructions.append(
                    f"Take {len(self.directions[i:j])} steps {self.direction_translation[self.directions[i]]}"
                )
            i, j = j, j + 1
        
        if j == len(self.directions):
            if len(self.directions[i:j]) == 1:
                translated_instructions.append(
                    f"Take 1 step {self.direction_translation[self.directions[i]]}"
                )
            else:
                translated_instructions.append(
                    f"Take {len(self.directions[i:j])} steps {self.direction_translation[self.directions[i]]}"
                )
        return "\n".join(translated_instructions)
```

You don't have to adopt the solution above, but you should remove at least one of the unnecessary loops in place now.  Also ensure your solution works in CW, it seems like it wouldn't at the moment?  I ask because when I suggested pulling somethings out you only partially did it. 

Consider the original code:
```python
                if len(self.directions[i:j]) == 1:
                    translated_instructions.append(
                        f"Take 1 step {self.direction_translation[self.directions[i]]}"
                    )
                    if j == len(self.directions):
                        break
                    i = j
                    j = i + 1
                else:
                    translated_instructions.append(
                        f"Take {len(self.directions[i:j])} steps {self.direction_translation[self.directions[i]]}"
                    )
                    if j == len(self.directions):
                        break
                    i = j
                    j = i + 1
```
is not the same functionality as you now have.
What I was suggesting was something like:
```python
                if len(self.directions[i:j]) == 1:
                    translated_instructions.append(
                        f"Take 1 step {self.direction_translation[self.directions[i]]}"
                    )
                else:
                    translated_instructions.append(
                        f"Take {len(self.directions[i:j])} steps {self.direction_translation[self.directions[i]]}"
                    )
                if j == len(self.directions):
                    break
                i = j
                j = i + 1
```
Note the `break` out block is not in either if statement since it's pulled out of both.

PR #98, opened 2023-01-08
CW order strings

aepaysinger:


code_challenges/code_wars/order_strings.py
```diff
@@ -0,0 +1,38 @@
...
+                if item[i].isdigit():
+                    self.position_in_string[position] = int(item[i]) - 1
+                    position += 1
+        return self.position_in_string
+
+    def build_updated_order_string(self):
+        placement = 0
+        ordered_string = ""
+        while len(ordered_string) < len(self.text):
+            for key, value in self.position_in_string.items():
```
W-Ely, line 20: `key, value` describe what they are not what values they contain.

W-Ely (commented):
I'd like to see you try to take self.split_text and pass it to sorted and get the right order.

like:
```python
    sorted_text = sorted(self.split_text, key=<a function you need to write>)
    return " ".join(sorted_text)

def <the function you need to write>(self, ....):
   ...
```

https://docs.python.org/3/howto/sorting.html

aepaysinger:
i got the code to run with what we did and updated the tests. i kept the comments because i wasn't sure if you wanted me to keep it or not

tests/code_wars/test_order_strings.py
```diff
@@ -0,0 +1,34 @@
...
+#     change_order.find_the_number()
+
+#     assert change_order.position_in_string == {0: 1, 1: 0, 2: 3, 3: 2}
+
+
+# def test_orderstrings_build_updated_order_strings():
+#     change_order = OrderStrings("is2 Thi1s T4est 3a")
+#     change_order.find_the_number()
+
+#     assert change_order.build_updated_order_string() == "Thi1s is2 3a T4est"
```
W-Ely, line 15: Little more clean up, or uncomment and use.

W-Ely (approved):

PR #101, opened 2023-01-12
advent wrapping presents

aepaysinger:


code_challenges/advent_of_code/wrapping_presents.py
```diff
@@ -0,0 +1,60 @@
...
+    return present_measurements
+
+
+def find_surface_area():
+    present_measurements = find_dimensions()
+    total_wrapping_paper = 0
+    for present in present_measurements:
+        a = int(present[0]) * int(present[1])
+        b = int(present[1]) * int(present[2])
+        c = int(present[2]) * int(present[0])
```
W-Ely, line 23: Anytime you find yourself doing present[0], present[1], ... Think if there is a better way to handle the variable.

```python
for length, width, height in present_measurements:
```

W-Ely (dismissed):

code_challenges/advent_of_code/wrapping_presents.py
aepaysinger: sounds like i need to really start thinking of writing code for others and not just me. someone else should be able to read it and understand what I am doing

W-Ely: > sounds like i need to really start thinking of writing code for others and not just me.

Always :o]

-------
The way you have changed this is not what I suggested and is less correct. 

This is literally what I meant:
```python
for length, width, height in present_measurements:
```

`int(present[0]) * int(present[1]) ` doesn't equal a length. It's a measure of area.

```python
for length, width, height in present_measurements:
    length_x_width = int(length) * int(width)
    width_x_height = int(width) * int(height)
    height_x_length = int(height) * int(length)
```

W-Ely (approved):

PR #103, opened 2023-01-13
CW camel case method

aepaysinger:


W-Ely (changes requested):
The code looks good. But now that you have a feel for how classes work you need to start thinking about how they would be used by someone else.  Let's say someone else can't see the code you have written, all they know is that they and instantiate the CamelCase class with a string.  Would it be intuitive that `camel_it.bring_words_together()` is the method they should run to get the output as camel case?  

So while these functions are named well for what they do, they are not named well for how they are expected to be used.

Another thought, you instantiate this class with a string of words.  This effectively make the class only good for one thing.  One point of a class is to be able to reuse the object.  It probably doesn't universally apply since you've been make the classes just to force practice with them, but in some cases it does, like this one.  A CamelCase class should be able to convert any string to camel case.

So the api would looks something like:
```python
camel_it.change_case("the string")
```

W-Ely:
> so would i want to put the function that does the final case changing (make_string_camel_case)(i changed the name) into the change case function? or vise versa? 

doesn't really make a difference, so long as the function is doing what it says and makes sense what it's doing.

> can have multiple functions that gets the thing done bur have the title of the final function make sense be the only one they need to use?

Also a-okay to have a single function that ties them together that a user would expect.

README.md
```diff
@@ -161,6 +161,12 @@
   * Module: `code_challenges/code_wars/change_it_up.py`
   * Tests: `tests/code_wars/test_change_it_up.py`
   * URL: [challenge url](https://www.codewars.com/kata/58039f8efca342e4f0000023/python)
+<<<<<<< HEAD
```
W-Ely, line 164: Always double check the files in github after a conflit resolution to ensure things look correct

W-Ely (changes requested):
Fix merge conflit resolution error

W-Ely (approved):

PR #104, opened 2023-01-16
Advent find easter bunny hq

aepaysinger:
for some reason i cant get the right distance. can you spot anything weird with it right off the bat? when you turn right or left your direction changes. N(0,0) was where i started so if R2 happens i turn 90 degress to the right which makes me face east and take 2 steps away from "home base"


code_challenges/advent_of_code/find_easter_bunny_hq.py
```diff
@@ -0,0 +1,56 @@
...
+    moves = elves_directions()
+    north_south = 0
+    east_west = 0
+    direction = "N"
+
+    moves = [move.strip() for move in moves]
+    print(moves)
+    for move in moves:
+        left_or_right = move[0]
+        steps = move[1]
```
W-Ely, line 19: What does line 19 do when it hits a number move like `L193`?

Instead of:
```python
    moves = [move.strip() for move in moves]
    print(moves)
    for move in moves:
        left_or_right = move[0]
        steps = move[1]   
```
I'd try:
```python
    for move in moves:
        move = move.strip()
        left_or_right, steps = move[1], int(move[1:])
        if ....
```

aepaysinger: i knew i wasnt seeing something!

code_challenges/advent_of_code/find_easter_bunny_hq.py
```diff
@@ -11,54 +11,233 @@ def find_easter_bunny_hq():
...
-
+    stops = set((0, 0))
+    double_stop = None
+    
+ 
     moves = [move.strip() for move in moves]
-    visited_locations = [(0,0)]
     for move in moves:
         left_or_right = move[0]
         steps = move[1:]
```
W-Ely, line 21:
```python
steps = int(move[1:])
```

code_challenges/advent_of_code/find_easter_bunny_hq.py
```diff
@@ -11,54 +11,233 @@ def find_easter_bunny_hq():
...
+                if (north_south, i) in stops:
+                    print(f"i have already been here {(north_south, i)}")
+                    double_stop = [north_south, i]
+                    if double_stop[0] < 0:
+                        double_stop[0] = double_stop[0] * -1
+                    elif double_stop[1] < 0:
+                        double_stop[1] = double_stop[1] * -1
+                    return double_stop[0] + double_stop[1]
+                else:
+                    stops.add((north_south, i))
```
W-Ely, line 138: What is a better way to handle code that needs to be repeated verbatim multiple times?

W-Ely (dismissed):

PR #105, opened 2023-01-20
advent chronal calibration

aepaysinger:
i got the first function working fine but now i have to find when the calibration is the same and that means i might have to start the loop at the beginning again but im not sure if im doing that right. i was trying to understand itertools.cycle but wasnt sure if that was the right fit


W-Ely:
> if that was the right fit?

Did it work?  If so not a terrible fit. 

[The docs](https://docs.python.org/3/library/itertools.html#itertools.cycle) say it looks something like:
```python
def cycle(iterable):
    saved = []
    for element in iterable:
        yield element
        saved.append(element)
    while saved:
        for element in saved:
              yield element
```

It's great to use the standard library, but the whole point of doing these exercises it to learn the syntax of the language. You should be able to write a semblance of the above without needing the built in library.  1) You should learn how yield works. 2) you don't really need a copy created, so why not something like:
```python
while True:
    for frequency in frequencies:
```

W-Ely (approved):

PR #106, opened 2023-01-28
CW equal sides of an array

aepaysinger:


code_challenges/code_wars/equal_sides_of_an_array.py
```diff
@@ -0,0 +1,34 @@
...
+        j = 2
+        for i in range(len(self.numbers)):
+            left_side = sum(self.numbers[: i + 1])
+            right_side = sum(self.numbers[j:])
+            if left_side == right_side:
+                return i + 1
+            j += 1
+            if j > len(self.numbers):
+                break
+        return -1
```
W-Ely, line 23: Good job keeping track of a couple pointers here, i and j.

The big O here is O(n^2), do you see why?  You're creating and summing multiple slices of the original list.

Compare the complexity to this solution:
```python
def find_even_index(numbers):
    right_total = sum(numbers)
    left_total = 0
    for i, value in enumerate(numbers):
        right_total -= value
        if right_total == left_total:
            return i
        left_total += value
    return -1
```

You don't have to change your solution here but if this were in an interview after your solution you would be ask about time complexity and also how you could improve it so it would be important to conceive of this way even if it weren't the choice you went with first.

W-Ely (dismissed):

code_challenges/code_wars/equal_sides_of_an_array.py
aepaysinger: i see. So this solution would need just a bit of tweaking but the point is you find the sum first and just subtract because that would just be one move vs summing it each time which would be multiple moves. One of those things that just seems instant but work is happening behind the scenes

W-Ely: Slicing is also not instant, a new list is created in memory each time.

The function above is complete without tweaking.  Which parts seem like they would need adjusting? Or which cases do you think might not work?

aepaysinger: you are right it doesn't need tweaking. because you subtract first then check then add.

aepaysinger: and because of this i dont have to do the two if statements at the start

PR #108, opened 2023-02-01
CW sudokue board validator

aepaysinger:
having issues with the test for the check_sections method. I ran into a recursion error. i know we have talked about this before but I can't remember how to handle the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,10 @@
   * Module: `code_challenges/code_wars/unique_in_order.py`
   * Tests: `tests/code_wars/test_unique_in_order.py`
   * URL: [challenge url](https://www.codewars.com/kata/54e6533c92449cc251001667/python)
+### Sudoku board validator (6kyu)
+  * Module: `code_challenges/code_wars/sudoku_board_validator.py`
+  * Tests: `test/code_wars/test_sudoku_board_validator.py`
+  * URL: [challenge url](https://www.codewars.com/kata/63d1bac72de941033dbf87ae/train/python)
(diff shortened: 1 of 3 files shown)


W-Ely:
> I ran into a recursion error.

What does `recursion` mean?  Show me the line(s) of code that recursively calls itself.

aepaysinger:
I call the function on line 59 but not sure why it is calling itself
multiple times

aepaysinger:
Nevermind! I just realized what I did. I didn't use the .notation so it's
just calling over and over. I'll fix it when I get home

W-Ely:
> but not sure why it is calling itself multiple times

If you don't see the function calling another function that calls itself, it isn't.

> Nevermind! I just realized what I did. I didn't use the .notation so it's
just calling over and over. I'll fix it when I get home

This is incorrect.

W-Ely:
```python
def test_sudoku_check_sections():
    # .........
    assert test_sudoku_check_sections() == False
```

`assert test_sudoku_check_sections() ...` is invoking which function?

aepaysinger:
It's calling test_sudokue_get_get sections but what I want is it to assert
sudoku_result.get_sections() == False

W-Ely:
> It's calling test_sudokue_get_get

Right so it is invoking itself, the test function's name is `test_sudoku_check_sections` so calling `test_sudoku_check_sections()` causes the test to call itself over and over.

W-Ely (approved):

PR #109, opened 2023-02-06
CW zip with

aepaysinger:


code_challenges/code_wars/zip_with.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ZipWith:
+    def __init__(self, func, array1, array2):
+        self.func = func
+        self.array1 = array1
+        self.array2 = array2
+
+    def zip_with(self):
+        return [self.func(a, b) for a, b in zip(self.array1, self.array2)]
```
W-Ely, line 11: Part of the purpose of classes is reusability.  This class should be instantiated with a zip function, but be able to zip any 2 arrays.

code_challenges/code_wars/zip_with.py
```diff
@@ -0,0 +1,24 @@
+from operator import add, sub, mul
```
W-Ely, line 1: unused imports should be removed.

W-Ely (changes requested):

code_challenges/code_wars/zip_with.py
aepaysinger: i have that there for my __if__ name == "__main__" block.

W-Ely: Are all three used in your if name == main block?  If not, remove the ones that are not used.  Unused imports should not be present.

W-Ely (approved):

W-Ely (approved):

PR #113, opened 2023-03-15
CW-cut rope

aepaysinger:


code_challenges/code_wars/cut_rope.py
```diff
@@ -0,0 +1,48 @@
+def cut_rope(length, m, n):
+    rope = []
+    different_lengths = {}
+    for _ in range(length):
+        rope.append("-")
```
W-Ely, line 5: Turn this into a list comprehension.

code_challenges/code_wars/cut_rope.py
```diff
@@ -0,0 +1,48 @@
...
+
+    count_m = 0
+    count_n = 0
+    for i, section in enumerate(rope):
+        if section == "-":
+            count_m += 1
+            count_n += 1
+        if count_m == m:
+            rope.insert(i + 1, ".")
```
W-Ely, line 14: You are modifying rope as you are iterating over it.  This isn't good practice, have to find a different way to do this.

W-Ely (changes requested):
You should never modify a collection as you iterate over it.

W-Ely (approved):

W-Ely (approved):

PR #114, opened 2023-03-20
CW pig

aepaysinger:


code_challenges/code_wars/pig_latin.py
```diff
@@ -0,0 +1,18 @@
+def pig_it(text):
+    text = text.split(" ")
+    translated_text = ""
+    punctuations = [" ", ".", "!", "?"]
```
W-Ely, line 4: This should be a set, tell my why?

code_challenges/code_wars/pig_latin.py
```diff
@@ -0,0 +1,18 @@
...
+    for word in text:
+        if word in punctuations:
+            translated_text += word
+        else:
+            translated_text += word[1:] + word[0] + "ay" + " "
+    if translated_text[-1] == " ":
+        return translated_text[:-1]
+    else:
```
W-Ely, line 12: Remove this unneeded else.

W-Ely (changes requested):

code_challenges/code_wars/pig_latin.py
aepaysinger: theres only one of each option and it is quicker

W-Ely: Why is it quicker?

aepaysinger: the run time is O(n) vs O(1). With the list it has to loop through the list every time. With the set it only needs to do it once then that information is stored

W-Ely: > With the set it only needs to do it once then that information is stored

This isn't correct.  See if you can good why a set is O(n) lookup.

W-Ely (approved):

PR #115, opened 2023-03-27
CW coordinates validator

aepaysinger:


W-Ely (changes requested):
See if you can find whats wrong here.

code_challenges/code_wars/coordinates_validator.py
```diff
@@ -0,0 +1,19 @@
+def is_valid_coordinates(coordinates):
+    for character in coordinates:
+        if character.isalpha():
+            return False
+    try:
+        x_coordinate = float(coordinates.split(", ")[0])
```
W-Ely, line 6: Only do this split once.

code_challenges/code_wars/coordinates_validator.py
```diff
@@ -0,0 +1,19 @@
...
+        y_coordinate = float(coordinates.split(", ")[1])
+    except:
```
W-Ely, line 8: No bare excepts, also write a test to catch whatever case this is handling.

code_challenges/code_wars/coordinates_validator.py
```diff
@@ -0,0 +1,19 @@
...
+        return False
+    if len(list(coordinates.split(","))) > 2:
```
W-Ely, line 10: split creates a list, no need to list it.

W-Ely (changes requested):

code_challenges/code_wars/coordinates_validator.py
aepaysinger: i belive i do with my test b

W-Ely: Great there is a test. Still `No bare excepts,`. You should (almost) never except _all_ exceptions, which one(s) are you expecting, except only it or them.

PR #118, opened 2023-03-29
Leet two sum

aepaysinger:


code_challenges/adhoc/two_sum.py
```diff
@@ -0,0 +1,14 @@
+def two_sum(numbers, target):
+
+    for i in range(len(numbers)):
+        for j in range(1, len(numbers)):
```
W-Ely, line 4: This is an O(n^2) solution.  Which is fine to come up with first.  Can you think of a way to do this in single loop over the list?

code_challenges/adhoc/two_sum.py
```diff
@@ -0,0 +1,18 @@
+def two_sum(numbers, target):
+    j = len(numbers) - 1
+    numbers.sort()
```
W-Ely, line 3: `numbers.sort()` This is more than a single loop. It O(n log n) I think.

code_challenges/adhoc/two_sum.py
```diff
@@ -0,0 +1,18 @@
...
+    for i in numbers:
```
W-Ely, line 4: And this is an additional single loop.

code_challenges/adhoc/two_sum.py
```diff
@@ -0,0 +1,18 @@
...
+        while numbers[i] + numbers[j] > target:
```
W-Ely, line 5: But this could also be a full loop couln't it?

imagine numbers = [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,....(on billion times then), 2,3] and target = 3.

W-Ely (commented):
How can you do it without sorting (which is more than a single loop) the list and only looping over the list once?

How might a set help you with this?

aepaysinger:
i don't think i can use a set since there can be two of the same numbers in the original list

W-Ely (approved):

W-Ely:
> i don't think i can use a set since there can be two of the same numbers in the original list

I see, it want's the index back not the numbers. If it wanted the numbers a set would work just fine.

Interesting that this branch is `leet` but it's being put in adhoc 😉

PR #121, opened 2023-04-14
rotate text

aepaysinger:


code_challenges/code_wars/rotate_text.py
```diff
@@ -0,0 +1,7 @@
+def rotate(str_):
+    rotated_words = []
+    word = str_
```
W-Ely, line 3: Why are you doing this line?

aepaysinger: i was thinking that i didn't want to change what i was looping over. originally i did enumerate(_str) and didn't want to change what i was looping over. but i changed it to loop over the range(len(_str)) which wouldnt be changing so i guess i dont need to do that now

code_challenges/code_wars/rotate_text.py
```diff
@@ -0,0 +1,7 @@
+def rotate(str_):
+    rotated_words = []
+
+    for _ in range(len(str_)):
+        rotated_words.append(str_[1:] + str_[0])
+        str_ = str_[1:] + str_[0]
```
W-Ely, line 6:
```suggestion
        str_ = str_[1:] + str_[0]
        rotated_words.append(str_)
```

PR #122, opened 2023-04-17
CW morse code

aepaysinger:


code_challenges/code_wars/morse_code.py
```diff
@@ -0,0 +1,75 @@
...
+    }
+    translation = ""
+    morse_code = morse_code.split(" ")
+
+    for i in range(len(morse_code) - 1, -1, -1):
+        if morse_code[i] == "":
+            continue
+        else:
+            morse_code = morse_code[: i + 1]
+            break
```
W-Ely, line 68: Is this needed?

W-Ely: I see, without some kind of handling line 71 would fail.

code_challenges/code_wars/morse_code.py
```diff
@@ -0,0 +1,74 @@
...
+    translation = ""
+    morse_code = morse_code.strip()
+    morse_code = morse_code.split(" ")
+
+    for i in range(len(morse_code)):
+        if morse_code[i] == "" and morse_code[i + 1] == "":
+            translation += " "
+        else:
+            translation += translator[morse_code[i]]
+    return translation
```
W-Ely, line 69: This probably works to pass the tests.  Now let's think about a real world situation.  The morse code operator is tapping out a message.  What is the chance that in between each letter is exactly 3 spaces?  How would you code fail? How could you fix it?

aepaysinger: i guess they could accidentally put more than 3 spaces and with my code that would add an extra space to the translation. i guess i could add and and statement and say and morse_code[i + 2] != ""

PR #128, opened 2023-05-22
advent password philosophy

aepaysinger:
for some reason i am not getting the right outcome for the check_password_validity_b() when i enter the results in advent. it says my amount is too high. can you spot where i might be adding more items then needed?


W-Ely:
> for some reason i am not getting the right outcome for the check_password_validity_b() when i enter the results in advent. it says my amount is too high. can you spot where i might be adding more items then needed?

What is b supposed to do?

W-Ely:
Your interpretation
> It checks to see if the letter is at either index. 

isn't correct, see this line
> Exactly one of these positions must contain the given letter.

and the third example.

aepaysinger:
i see what i missed now. thanks!! quick fix

W-Ely (approved):

PR #130, opened 2023-05-25
advent dive 

aepaysinger:


code_challenges/advent_of_code/dive.py
```diff
@@ -0,0 +1,41 @@
...
+    with open("./code_challenges/advent_of_code/directions_input") as input:
+        directions = input.read().split("\n")
+    return [direction.split(" ") for direction in directions]
+
+
+def find_depth():
+    directions = get_directions()
+    horizontal = 0
+    depth = 0
+    for direction in directions:
```
W-Ely, line 11: Expand direction out here into variable that have meaning instead of using direction[0] and direction[1] below.

W-Ely (approved):

PR #132, opened 2023-05-30
Advent giant squid

aepaysinger:


code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+
+    for number_to_call in numbers_to_call:
+        for board in boards:
+            for row_number, rows in enumerate(boards[board]):
+                for i, number in enumerate(rows):
+                    if number_to_call == number:
+                        boards[board][row_number][i] = "X"
+                    if boards[board][row_number] == ["X", "X", "X", "X", "X"]:
+                        return boards[board], number
+                    for _ in range(len(rows)):
```
W-Ely, line 34: `_` is only acceptable if the variable will not be used.

W-Ely: Line 29 is already looping over the rows, why is this for loop needed?

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+
+    return final_boards, numbers_to_call
+
+
+def play_bingo():
+    boards, numbers_to_call = get_bingo_info()
+
+    for number_to_call in numbers_to_call:
+        for board in boards:
+            for row_number, rows in enumerate(boards[board]):
```
W-Ely, line 28: Why not `for row_number, rows in enumerate(board):`?

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+                for i, number in enumerate(rows):
+                    if number_to_call == number:
+                        boards[board][row_number][i] = "X"
```
W-Ely, line 31: Would this work instead? `boards[board][row_number][i] = "X"` --> `rows[i] = "X"`

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+    return final_boards, numbers_to_call
+
+
+def play_bingo():
+    boards, numbers_to_call = get_bingo_info()
+
+    for number_to_call in numbers_to_call:
+        for board in boards:
+            for row_number, rows in enumerate(boards[board]):
+                for i, number in enumerate(rows):
```
W-Ely, line 29: I would expect a variable called `rows` to contain multiple `row`?  Something isn't named correctly in this.

Maybe this line `for row_number, rows in enumerate(boards[board]):` should actually read `for row_number, row in enumerate(board):` note singular `row`

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+                            and boards[board][1][_] == "X"
+                            and boards[board][2][_] == "X"
+                            and boards[board][3][_] == "X"
+                            and boards[board][4][_] == "X"
+                        ):
+                            return boards[board], number
+
+
+def find_winner_score():
+    wining_board, winning_number = play_bingo()
```
W-Ely, line 46: nit: spelling

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+    score = 0
+    for rows in wining_board:
```
W-Ely, line 48: when you iterate over this board do you get multiple rows at a time or only one?

W-Ely (commented):
What about diagonal wins?

aepaysinger:
> What about diagonal wins?

for this challenge it didn't include diagonal wins

code_challenges/advent_of_code/giant_squid.py
aepaysinger: i want to check right after the x is inserted into the board

W-Ely: I still do see why you need the loop?  Wouldn't `i` be the only location that might have changed?  

If you removed the loop but kept the if statement and replaced `_` with `i` would it do the same thing?  Just not waste cycles.

code_challenges/advent_of_code/giant_squid.py
```diff
@@ -0,0 +1,93 @@
...
+        final_boards[board_number] = final_boards.get(board_number, []) + [numbers]
+
+    return final_boards, numbers_to_call
+
+
+def play_bingo():
+    boards, numbers_to_call = get_bingo_info()
+
+    for number_to_call in numbers_to_call:
+        for board in boards:
```
W-Ely, line 27: What about something like `for board_number, board in boards.items():`?  So you don't have to have boards[board] everywhere.

code_challenges/advent_of_code/giant_squid.py
aepaysinger: that is better. i changed variable names a lot when putting this together.

PR #133, opened 2023-06-07
advent hydrothermal venture

aepaysinger:
so i know i can pull some of the same info from bother functions into another function so i dont write out things twice.
i passed the first test but the answer i got for the second function is too low. not sure what im missing. when i checked them via pytest coverage it looks like everything is being covered


W-Ely (commented):
I would break this problem down as
1) format the data into tuple pairs like `[((0, 9), (9, 0)), ....]`
2) iterate over these pairs and find all the points between. This would take a little thought. but the result would look like (example shows only one coord pair, but all should be in the output:
coord pairs in: `[((0, 9), (9, 0))]`
points out: `[(0, 9), (1, 8), (2, 7), (3, 6), (4, 5), (5, 4), (6, 3), (7, 2), (8, 1), (9, 0)]`
3) from here (or even in the same loop, populate a frequency map of all the points from between all the coordinate pairs (only one pair exampled - explains the count always being one. The real one would have overlaps so should have 2s and greater:
 points in: `[(0, 9), (1, 8), (2, 7), (3, 6), (4, 5), (5, 4), (6, 3), (7, 2), (8, 1), (9, 0)]`
freq map out: `{(0, 9): 1,  (1, 8): 1, (2, 7): 1, (3, 6): 1,  (4, 5): 1, (5, 4): 1, (6, 3): 1, (7, 2): 1, (8, 1): 1, (9, 0): 1}`
4: count the values in the map greater than 2. (could also be done in the same loop is so inclinded.

code_challenges/advent_of_code/hydrothermal_venture.py
```diff
@@ -0,0 +1,138 @@
...
+    with open("code_challenges/advent_of_code/vents_input") as vents_input:
+        vents_coordinates = vents_input.read().split("\n")
+    # for vent_coordinate in vents_coordinates:
+    #     vent_coordinate.split("->")
+        
+    vents_coordinates = [vent_coordinate.split("->") for vent_coordinate in vents_coordinates]
+    vents_coordinates = [coordinate.strip() for vent_coordinate in vents_coordinates for coordinate in vent_coordinate]
+    vents_coordinates = [vent_coordinate.split(",") for vent_coordinate in vents_coordinates]
+    # vents_coordinates = [int(coordinate) for vent_coordinate in vents_coordinates for coordinate in vent_coordinate]
+    return vents_coordinates
```
aepaysinger, line 11: it is my understanding that i am getting a list of pairs. i just have them in a list not a tuple. does that matter?
[['0', '9'], ['5', '9'], ['8', '0'], ['0', '8'], ['9', '4'], ['3', '4'], ['2', '2'], ['2', '1'], ['7', '0'], ['7', '4'], ['6', '4'], ['2', '0'], ['0', '9'], ['2', '9'], ['3', '4'], ['1', '4'], ['0', '0'], ['8', '8'], ['5', '5'], ['8', '2']]

W-Ely: This is a list of coordinates, not coordinate pairs.  The start and end point should be grouped together into a pair.

[(['0', '9'], ['5', '9']), (['8', '0'], ['0', '8']), (['9', '4'], ['3', '4']), .......]

W-Ely: But also, if you don't need the properties of a list, you should use a tuple.

aepaysinger: oops you are right. i put them into pairs at the start of the function but could do that in the in its own function since i do it twice

tests/advent_of_code/test_hydrothermal_venture.py
```diff
@@ -0,0 +1,122 @@
...
+            [3, 4],
+            [1, 4],
+            [0, 0],
+            [8, 8],
+            [5, 5],
+            [8, 2],
+        ],
+        9,
+        9,
+    )
```
W-Ely, line 120:
```suggestion
    expected = (
        [
            ([0, 9], [5, 9]),
            ([8, 0], [0, 8]),
            ([9, 4], [3, 4]),
            ([2, 2], [2, 1]),
            ([7, 0], [7, 4]),
            ([6, 4], [2, 0]),
            ([0, 9], [2, 9]),
            ([3, 4], [1, 4]),
            ([0, 0], [8, 8]),
            ([5, 5], [8, 2]),
        ],
        9,
        9,
    )
```
This roughly what the shape of the response should be.

code_challenges/advent_of_code/hydrothermal_venture.py
W-Ely: They are still not in pairs.  Your test for `get_coordinate_pairs` shows it.  It returns a flat list of coordinates, not a list of pair of coordinates. 

Also you should mock the call to `open` since you have so much logic in `get_vents_input`, then you can test it too.

code_challenges/advent_of_code/hydrothermal_venture.py
```diff
@@ -0,0 +1,145 @@
...
+    return coordinate_pairs, biggest_x, biggest_y
+
+
+def mark_coordinates_horizontal_vertical():
+    coordinate_pairs, biggest_x, biggest_y = get_coordinate_pairs()
+    coordinates_map = [[0 for i in range(biggest_x + 1)] for i in range(biggest_y + 1)]
+    x = 0
+    y = 1
+
+    for coordinates in coordinate_pairs:
```
W-Ely, line 56:
```
In [1]: a = [(("a", "b"), ("c", "d")), (("e", "f"), ("g", "h"))]

In [2]: for (x1, y1), (x2, y2) in a:
   ...:     print(x1, y1, x2, y2)
   ...:
a b c d
e f g h

In [3]:
```

aepaysinger: so i think i was trying to do something like that but because it had the 2 sets i was thinking you cant do it. this reads WAY better

aepaysinger:
after our session i made the function count_coordinate_point(). this should do what we discussed. the issue i have found is when the points 7,0 -> 7,4 and 5,5 -> 8,2 are entered its not working correctly. point (7,3) shows up in both when i run the numbers alone but if i run both of them { (7,3) : 1} is what i get instead of {(7,3): 2} and im not sure why. all the other points are working

code_challenges/advent_of_code/hydrothermal_venture.py
```diff
@@ -0,0 +1,186 @@
...
+                    coordinate_points[(x1, coordinate)] = coordinate_points.get((x1, coordinate), 0) + 1
+        elif x1 < x2 and y1 < y2:
+            x_coordinate = x1
+            for coordinate in range(x1, x2 +1):
+                coordinate_points[(coordinate, x_coordinate)] = coordinate_points.get((coordinate, x_coordinate), 0) + 1
+                x_coordinate += 1
+        elif x1 < x2 and y1 > y2:
+            x_coordinate = x2
+            for coordinate in range(y2, y1 + 1):
+                coordinate_points[(x_coordinate, coordinate)] = coordinate_points.get((coordinate, x_coordinate), 0) + 1
```
W-Ely, line 67: This line is incorrect I think.

aepaysinger: oh i see it!!! well i fixed it but i am still getting an answer that is too low for the advent challenge so i am going through and cleaning it all up so i can write more tests and hopefully see whats going on

PR #137, opened 2023-07-05
CW skills master

aepaysinger:


code_challenges/code_wars/skills_master.py
```diff
@@ -0,0 +1,15 @@
+def count_skills(tree, required):
+    if required == set():
+        return 0
+
+    visited_levels = set()
+    for number in required:
+        level = tree[number]
+        while level not in visited_levels and level not in required:
```
W-Ely, line 8: `and level not in required` nice, this makes it much much faster.

W-Ely (approved):

PR #139, opened 2023-07-09
cw expanded form

aepaysinger:


code_challenges/code_wars/expanded_form.py
```diff
@@ -0,0 +1,12 @@
+def expanded_form(num):
+    num = str(num)
+    expanded = ""
+    i = 1
+    for character in num:
```
W-Ely, line 5: use enumerate instead.

aepaysinger: i didn't realize you could set a starting point for enumerate. that was the only reason i didnt use it. i have made the changes now.

W-Ely (approved):

PR #142, opened 2023-08-02
Advent chess game

aepaysinger:


code_challenges/advent_of_code/chess_game.py
```diff
@@ -0,0 +1,30 @@
...
+    return password
+
+
+def find_second_pass_code(code):
+    password = ["-" for _ in range(8)]
+    number = 0
+    while "-" in password:
+        test_hash = hashlib.md5(code + str(number).encode("utf-8")).hexdigest()
+        if test_hash.startswith("00000"):
+            if test_hash[5].isdigit():
```
W-Ely, line 24: Combine these 2.

code_challenges/advent_of_code/chess_game.py
```diff
@@ -0,0 +1,30 @@
+import hashlib
+
+
+def find_pass_code(code):
+    used_hashes = []
+    number = 0
+    password = ""
+    while len(used_hashes) < 8:
```
W-Ely, line 8: is this while loop correct?  You never need more than 7?

code_challenges/advent_of_code/chess_game.py
```diff
@@ -0,0 +1,30 @@
...
+        test_hash = hashlib.md5(code + str(number).encode("utf-8")).hexdigest()
+        if test_hash.startswith("00000"):
+            used_hashes.append(test_hash)
+        number += 1
+    for hash in used_hashes:
+        password += hash[5]
```
W-Ely, line 14: why not do this on line 11 instead?

code_challenges/advent_of_code/chess_game.py
aepaysinger: doesnt this always give me 8? because when the len(used_hashes) == 7 it enters the block and then adds to used_hashes and then the len is 8 and it doesnt enter

W-Ely: Sure, I was off by one, the question was meant to ask if you need a fixed number. I you only ever need 8 okay.

W-Ely (approved):

PR #145, opened 2023-08-15
Advent get past gaurd

aepaysinger:


code_challenges/advent_of_code/get_past_gaurd.py
```diff
@@ -0,0 +1,107 @@
...
+def get_gaurd_schedule():
+    with open("code_challenges/advent_of_code/gaurd_schedule") as shifts:
+        gaurd_schedule = shifts.read().split("\n")
+
+    return [shift.replace("]", "").replace("[", "") for shift in gaurd_schedule]
+
+
+def organize_schedule():
+    schedule = get_gaurd_schedule()
+    schedule = [[shift[0:16], shift[17:]] for shift in schedule]
```
W-Ely, line 13: I would like you to start thinking about whether you need a list, or if a tuple would work.

For example, does this work downstream in your code?
```suggestion
    schedule = [(shift[0:16], shift[17:]) for shift in schedule]
```

aepaysinger: okay. i know you have mentioned it a few times. its a hard habit to break. it does work and i updated the function as well as the test for that function

W-Ely (approved):

PR #146, opened 2023-08-17
CW compare substring values 

aepaysinger:


code_challenges/code_wars/compare_substring_values.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    highest_substring_value = 0
+    substring_value = 0
+    for strng in strngs:
+        for character in strng:
+            substring_value += alphabet[character]
+        if substring_value > highest_substring_value:
+            highest_substring_value = substring_value
+        substring_value = 0
```
W-Ely, line 38:
```suggestion
    for string in strings:
        substring_value = 0
        for character in string:
            substring_value += alphabet[character]
        if substring_value > highest_substring_value:
            highest_substring_value = substring_value
```

W-Ely (dismissed):
strng --> string thoughout.  You wouldn't want to name a variable after a type, but `str` is the type in python so `string` is safe to use

W-Ely (approved):

PR #147, opened 2023-08-21
CW amount of sets

aepaysinger:


aepaysinger:
in order for this to pass in codewars i need a faster run time but im not sure how to do that. any suggestions?

W-Ely:
Do you actually need to create the sets in order to count them/know how many there will be?  

Consider this, if we compare start set lengths (already deduplicated) to the combinations possible it looks something like:
```
length --> count
1 --> 1
2 --> 3
3 --> 7
4 --> 15
5 ....
```
From this it isn't really clear whats going on.  Certainly enough to try a few things, but we can clear this up a little by not looking at the final count, but instead the amount that each extra value in the set adds to the total count.
```
length --> additional sets more than the last.
1 --> 1
2 --> 2 
3 --> 4
4 --> 8
5 --> 16
6 --> 32
7 --> 64
8 --> 128
9 --> 256
...
```
With the above it is very clear what algorithm needs to be used to calculate the number without ever creating a single set.  All you need is the length.

aepaysinger:
what a great idea! i don't need to know what the sets are!

W-Ely (approved):

PR #148, opened 2023-08-28
CW longest collazt sequence

aepaysinger:


code_challenges/code_wars/longest_collatz_sequence.py
```diff
@@ -21,7 +21,7 @@ def find_longest_length(numbers):
     winner = 0
     winner_length = 0
     for number in numbers:
-        if len(numbers_with_sequence[number]) > winner_length:
+        if numbers_with_sequence[number] > winner_length:
             winner = number
-            winner_length = len(numbers_with_sequence[number])
+            winner_length = numbers_with_sequence[number]
```
W-Ely, line 26: Use the max function on `numbers_with_sequence` instead.  Note you'll have to use the `key` kwarg.

aepaysinger: can i use max since? if there are 2 numbers with the same length then it goes based off of which one is first in the list and i thought dictionaries dont always keep the same order

W-Ely: > can i use max since? if there are 2 numbers with the same length then it goes based off of which one is first in the list

Your right, _since you are using a dictionary when you don't need one_, you can't be sure of the order. (even though in Python 3 you can...)

code_challenges/code_wars/longest_collatz_sequence.py
```diff
@@ -0,0 +1,22 @@
...
+    if num % 2 == 0:
+        return num / 2
+    else:
+        return 3 * num + 1
+
+
+def find_longest_length(numbers):
+    return max(
+        find_collatz_sequence(numbers), key=lambda x: find_collatz_sequence(numbers)[x]
+    )
```
W-Ely, line 22: You shouldn't call  `find_collatz_sequence` twice here.
```suggestion
    numbers_with_sequence =  find_collatz_sequence(numbers)
    return max(
        numbers_with_sequence, key=lambda x: numbers_with_sequence[x]
    )
```

aepaysinger: i had it that way and then thought "i shouldnt make a variable when i dont need to." But I see how my original thought to have the function call in a variable is better.

aepaysinger:
would it be better to do a list of tuples?

W-Ely (approved):
Can't use tuples because Tuples are immutable. Nice thought though, always consider a Tuple before a list and choose a list if a tuple won't work.

PR #150, opened 2023-08-30
Advent nice strings

aepaysinger:


code_challenges/advent_of_code/nice_strings.py
```diff
@@ -0,0 +1,98 @@
...
+def find_the_pattern(text):
+    j = 2
+    for i in range(len(text)):
+        if i == len(text) - 2:
+            return False
+        elif text[i] == text[j]:
+            
+            return True
+        j += 1
+
```
aepaysinger, line 72: something isn't working right with find_double_pair and find_the_pattern because I am not able to get the correct answer for advent.......here are the instructions:
Realizing the error of his ways, Santa has switched to a better model of determining whether a string is naughty or nice. None of the old rules apply, as they are all clearly ridiculous.

Now, a nice string is one with all of the following properties:

It contains a pair of any two letters that appears at least twice in the string without overlapping, like xyxy (xy) or aabcdefgaa (aa), but not like aaa (aa, but it overlaps).
It contains at least one letter which repeats with exactly one letter between them, like xyx, abcdefeghi (efe), or even aaa.
For example:

qjhvhtzxzqqjkmpb is nice because is has a pair that appears twice (qj) and a letter that repeats with exactly one letter between them (zxz).
xxyxx is nice because it has a pair that appears twice and a letter that repeats with one between, even though the letters used by each rule overlap.
uurcxstgmygtbstg is naughty because it has a pair (tg) but no repeat with a single letter between them.
ieodomkazucvgmuy is naughty because it has a repeating letter with one between (odo), but no pair that appears twice.
How many strings are nice under these new rules?

PR #152, opened 2023-09-03
Advent repetition code

aepaysinger:


code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,50 @@
+def get_messages():
+    with open("./code_challenges/advent_of_code/message_input") as message_input:
+        messages = message_input.read().split("\n")
+    return messages
+
+
+def break_up_messages():
+    messages = get_messages()
+    character_breakdown = {}
```
W-Ely, line 9: You never use this dictionary as a dictionary, why not a list of character count dicts?

code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,50 @@
...
+                character_breakdown[f"position {str(i)}"].get(character, 0) + 1
+            )
+    return character_breakdown
+
+
+def find_hidden_message_high():
+    character_breakdown = break_up_messages()
+
+    hidden_message = ""
+    for position, characters in character_breakdown.items():
```
W-Ely, line 25: position isn't used because you are using a dictionary as a list.

W-Ely: Also, when you iterate over this dictionary, you are relying on it being in a certain order.  This isn't a dictionary thing even though it's working in this case.

code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def find_hidden_message_low():
+    character_breakdown = break_up_messages()
+
+    hidden_message = ""
+    for characters in character_breakdown.values():
+        lowest_character = None
+        lowest_count = float("inf")
+        for character, count in characters.items():
```
W-Ely, line 44: why not the `min` function.

code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def find_hidden_message_high():
+    character_breakdown = break_up_messages()
+
+    hidden_message = ""
+    for position, characters in character_breakdown.items():
+        highest_character = None
+        highest_count = 0
+        for character, count in characters.items():
```
W-Ely, line 28: What about the `max` function?

code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,38 @@
...
+        messages = message_input.read().split("\n")
+    return messages
+
+
+def break_up_messages():
+    messages = get_messages()
+    character_breakdown = []
+
+    for i in range(len(messages[0])):
+        character_breakdown.append({})
```
W-Ely, line 12: Is this loop needed?  Just create and append in the loop below.

code_challenges/advent_of_code/repetition_code.py
```diff
@@ -0,0 +1,37 @@
...
+    character_breakdown = []
+
+    for message in messages:
+        for i, character in enumerate(message):
+            while len(character_breakdown) < len(messages[0]):
+                character_breakdown.append({})
+            character_breakdown[i][character] = (
+                character_breakdown[i].get(character, 0) + 1
+            )
+    return character_breakdown
```
W-Ely, line 18: What I had in mind was something like:
```python
    for message in messages:
        character_counts = {}
        for character in message:
            character_counts[character] = character_counts.get(character, 0) + 1
```

But now I see that the number of dictionary is supposed equal the length of a single message. hmmm.  The way you had it before is better.

or consider:
```python
character_breakdown = []
for i in len(messages[0]:
    char_count = {}
    for message in messages:
        character = message[i]
        char_count[character] = char_count[character].get(character, 0) + 1
    character_breakdown.append(char_count)
```

W-Ely (approved):

PR #156, opened 2023-09-13
lonliest character 

aepaysinger:
O(log n)?


aepaysinger:
i would like to change my answer. I think it is O(n). No matter the size of the input it does one operation for each item in both loops.

W-Ely (approved):

PR #157, opened 2023-09-15
phone text

aepaysinger:
first i wanted to say O(n^2) because of the nested for loop. but since you dont have to loop through the whole dictionary every time i am thinking it is O(n).

### DIFF
--- a/code_challenges/code_wars/phone_text.py
+++ b/code_challenges/code_wars/phone_text.py
@@ -1,0 +1,14 @@
+def presses(phrase):
+    key_pad_touches = {
+        1: {"A", "D", "G", "J", "M", "P", "T", "W", " ", "*", "#", "1"},
+        2: {"B", "E", "H", "K", "N", "Q", "U", "X", "0"},
+        3: {"C", "F", "I", "L", "O", "R", "V", "Y"},
+        4: {"2", "3", "4", "5", "6", "S", "8", "Z"},
+        5: {"7", "9"},
+    }
+    number_of_touches = 0
+    for character in phrase.upper():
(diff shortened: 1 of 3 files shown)


W-Ely:
> first i wanted to say O(n^2) because of the nested for loop. but since you dont have to loop through the whole dictionary every time i am thinking it is O(n).

O(n) is correct, but has nothing to do with not needing to loop through the whole dictionary.  The Big O is always reflective of the worst case. Meaning if the input phrase in this case was "99999999999999999999999999..."

code_challenges/code_wars/phone_text.py
```diff
@@ -0,0 +1,14 @@
...
+        3: {"C", "F", "I", "L", "O", "R", "V", "Y"},
+        4: {"2", "3", "4", "5", "6", "S", "8", "Z"},
+        5: {"7", "9"},
+    }
+    number_of_touches = 0
+    for character in phrase.upper():
+        for presses, characters in key_pad_touches.items():
+            if character in characters:
+                number_of_touches += presses
+    return number_of_touches
```
W-Ely, line 14: Don't change it, but if someone said this needs to be at least 5x faster what would you suggest?

W-Ely (approved):

code_challenges/code_wars/phone_text.py
aepaysinger: change the dict. to have the key be each letter/number. then have the value be the amount of taps. 
for character in phrase.upper():
    number_of_touches += key_pad_touches[character]

???

aepaysinger: YES!

PR #158, opened 2023-09-18
CW-kebabize 

aepaysinger:
big O: O(1) : The function has one task

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,10 @@
   * Module: `code_challenges/code_wars/phone_text.py`
   * Tests: `tests/code_wars/test+phone_text.py`
   * URL: [Challenge URL](https://www.codewars.com/kata/54a2e93b22d236498400134b/python)
+### Kebabize (6kyu)
+  * Module: `code_challenges/code_wars/kebabize.py`
+  * Tests: `tests/code_wars/test_kebabize.py`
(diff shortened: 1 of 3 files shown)


W-Ely:
> big O: O(1) : The function has one task

This is incorrect, tell me why.

aepaysinger:
big O = O(n)
not O(1), it goes off of worst case. with the if else blocks its not only doing one thing.

W-Ely:
> big O = O(n)

and
> it goes off of worst case.

^-- yes

>  with the if else blocks its not only doing one thing.

Has nothing to do with the if/else block

Why is this O(n).

W-Ely:
> Linear time: 1 series operation for each item that is passed in.

What item(s) are being passed in?  Is it one thing or many?  How are you treating what is being passed in.

aepaysinger:
a single string is being passed in. one item is being passed in. then each character of the string goes through the for loop.

W-Ely:
> a single string

I think that this isn't correct. While technically maybe.  You are using this string as a collection of characters.  A string isn't one thing. It's a container of bytes.

So it's O(n) because as the string gets longer (containing more characters) the duration increases linearly.

aepaysinger:
okay so it is O(n) and it is this because as the input size gets bigger the run time increases and it is also doing 1 series of operations?

W-Ely (approved):

PR #159, opened 2023-09-19
cw not prime number complete

aepaysinger:
O(n)


W-Ely (approved):
O(n) not sure, but think it isn't important.  The complexity goes up  with both n (the range increasing) and the size of the numbers in the range..

PR #161, opened 2023-10-02
cw multiples of three

aepaysinger:
hasnt passed code wars due to run time

### DIFF
--- a/code_challenges/code_wars/multiples_of_three.py
+++ b/code_challenges/code_wars/multiples_of_three.py
@@ -1,0 +1,26 @@
+from itertools import permutations
+
+
+def find_mult_3(num):
+    multiples = set()
+    multiples_of_3 = set()
+    i = 1
+    num = str(num)
+    nums = [digit for digit in num]
+    while i <= len(str(num)):
+        perm = permutations(nums, i)
+        for j in list(perm):
+            multiples.add(j)
+        i += 1
+
+    for numbers in multiples:
+        new_num = ""
+        for number in numbers:
+            new_num += number
+        new_num = int(new_num)
+        if new_num == 0:
+            continue
+        if new_num % 3 == 0:
+            multiples_of_3.add(new_num)
+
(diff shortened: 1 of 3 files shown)


code_challenges/code_wars/multiples_of_three.py
```diff
@@ -0,0 +1,26 @@
...
+        new_num = ""
+        for number in numbers:
+            new_num += number
+        new_num = int(new_num)
+        if new_num == 0:
+            continue
+        if new_num % 3 == 0:
+            multiples_of_3.add(new_num)
+
+    return [len(multiples_of_3), max(multiples_of_3)]
```
W-Ely, line 26: I think you're doing just a little to much work.  This works and is basically the same approach:
```
from itertools import permutations


def find_mult_3(num):
    digits = str(num)
    max_multiple = 0
    count = 0
    seen = set()
    for i in range(len(digits)):
        for perm in permutations(digits, i + 1):
            multiple = int("".join(perm))
            if not multiple or multiple in seen:
                continue
            seen.add(multiple)
            if multiple % 3 == 0:
                max_multiple = max(max_multiple, multiple)
                count += 1
    return [count, max_multiple]
```

I did also try it without permutations:
```
from collections import Counter


def find_max_multiple(digits, counts):
    number = int("".join(sorted([digit for digit in digits], reverse=True)))
    while number % 3 != 0:
        number -= 1
    while not Counter(str(number)) <= counts:
         number -= 3
    return number


def find_mult_3(num):
    digits = str(num)
    counts = Counter(digits)
    max_multiple = find_max_multiple(digits, counts)
    count = 0
    for multiple in range(3, max_multiple + 1, 3):
        if Counter(str(multiple)) <= counts:
            count += 1
    return [count, max_multiple]
```
But it is too slow also.

W-Ely (dismissed):

W-Ely (approved):

PR #165, opened 2023-10-30
CW almost even

aepaysinger:


code_challenges/code_wars/almost_even.py
```diff
@@ -0,0 +1,10 @@
+def split_integer(num, parts):
+    integer_parts = []
+    part = num // parts
+    for _ in range(parts):
+        integer_parts.append(part)
+    i = len(integer_parts) - 1
+    for _ in range((num % parts)):
```
W-Ely, line 7:
```suggestion
    for _ in range(num % parts):
```

code_challenges/code_wars/almost_even.py
```diff
@@ -0,0 +1,10 @@
+def split_integer(num, parts):
+    integer_parts = []
+    part = num // parts
+    for _ in range(parts):
+        integer_parts.append(part)
```
W-Ely, line 5: Turn this into a list comprehension.

W-Ely (approved):

